PR #32, opened 2017-09-19
adding an edit in GitHub button to posts pages and events

jeremyzilar:


thisisdano:
Prettier than I expected!

PR #37, opened 2017-09-20
adding a new author file

SoncerrayBolling:


thisisdano:
We should be merging this into `demo` instead

PR #42, opened 2017-09-21
Use original src when provided with full url as src

thisisdano:


thisisdano:
So, if you have a link to an image file outside of S3, just put the full URL in the `src` param.
like {{< img src="https://domain.tld/path/to/img.jpg " >}} will just point to that specific path. 

{{< img src="image-uid " >}} will continue to point to S3 as usual!

PR #52, opened 2017-09-27
Create new event

jthalls:
@thisisdano can you please review this new event


thisisdano:
@jthalls This event is coded as mixed, but doesn't include any physical location info. That's OK, but we might want to consider adding it.

@jeremyzilar this is kinda what threw the errors — I made all the physical location metadata conditional to prevent this in the future

PR #60, opened 2017-09-29
Fixing broken links, building a better linker, and also a better devmode

thisisdano:
This PR fixes a bunch of old dead links remaining from the original migration. But it also does a couple other key things:

**There's now a new way to do dev mode**

Now, there's no need to modify the `config.yml` file. I made a new, dev mode config file called `config_dev.yml` which you access when running a hugo `serve`. 

Now, use `hugo serve --config=config_dev.yml`

This still ignores all posts prior to 2017, but it also includes a `.Site.Params` flag that we can access in the `link` shortcode that instructs the shortcode to ignore links to posts prior to 2017. This should remove all those pesky errors that make dev mode less than ideal.

**One link shortcode to rule them all**

Use `{{< link "thing-to-be-linked" >}}` for all internal links (it also work for external links)
- `{{< link "some-file-name.md" >}}` will link to a specific file, regardless of it place in the filesystem
- `{{< link "path/to/a/location" >}}` will link to a specific location from the web root, like `/tag/ux`
- `{{< link "http[s]://some.tld/path/to/file" >}}` will process the link as is


jeremyzilar:
Seems like there is a conflict with the `config.yml` file?

jeremyzilar:
I just switched to this branch and `hugo serve --config=config_dev.yml` worked for me, super fast!

PR #76, opened 2017-10-06
Adding in Issue templates

jeremyzilar:
### Problem Statement

We need a bit of guidance around how to use Issues and Pull Requests.
Maybe adding in some Issue templates is a good way to assist the workflow.

### Additional information
This was originally forked from the USWDS templates and modified.
https://github.com/18F/web-design-standards/tree/develop/.github

Here is the GitHub documentation on the files and how they work:
https://help.github.com/articles/creating-an-issue-template-for-your-repository/

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,9 @@
+### Define the problem
+A high-level description of the problem you are trying to solve.
+What did you expect to happen? What happened instead?
+
+### Additional information [optional]
+
+* Screen shot images
+* Notes
+* And so on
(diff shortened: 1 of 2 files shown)


thisisdano:
I don't think these templates have proven useful for USWDS. From what I've seen, the first step in any issue or PR is select all and delete. I wonder if we want to add anything at all, and if we do, if we should be really minimal and less prescriptive.

A minimal thing would be like `Tell us about more about what you'd like to see fixed or improved. Be explicit and include examples when possible. Thanks for contributing!`

jeremyzilar:
Good point. Closing

PR #78, opened 2017-10-08
Preparing Events

jeremyzilar:
The goal of this is to prepare the events pages to go live.
I am cleaning up the test events, and making sure that all the existing events are in working order.


jeremyzilar:
Here is the preview: https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/Event-Pages-Prep/

Issue #129, opened 2017-10-26
Add USWDS "An official gov't website" banner


ToniBonittoGSA:
Best example at https://www.usa.gov/ 

18f.gov and many others are missing the newer "Here's how you know" link that expands the banner

thisisdano:
Can anyone think of any reason why we wouldn't want to do this? Anything that we'd want to do before this?

I think it'd be relatively simple to implement.

AndreaSigz:
Nope. It would be great to get this on DigitalGov

thisisdano:
OK, thanks! I'll assign myself and we'll get it done!

thisisdano:
Now, to see how well the USWDS plays with our site...

ToniBonittoGSA:
Howdy... 
I just noticed that the space after the w in know (between know and the down arrow) is linked; it should not be. The same problem is on USA.gov so maybe it's universally incorrect. 
Who can fix that? See attached screen cap 

![screen shot 2017-11-07 at 12 15 06 pm 2](https://user-images.githubusercontent.com/19961574/32507517-a243bdcc-c3b5-11e7-8d5c-32d83141bbe3.png)

ToniBonittoGSA:
and 
![screen shot 2017-11-07 at 12 20 53 pm 2](https://user-images.githubusercontent.com/19961574/32507664-14d42606-c3b6-11e7-8f2f-71488a858b5b.png)

jeremyzilar:
This has been completed 🎉👾

PR #182, opened 2017-11-06
Add USWDS banner

thisisdano:
✅ **Ready for review**

This adds the USWDS banner to DigitalGov. It adds a number of files and dependencies.
- A gulp process
- USWDS 1.4.2
- A number of scss files to override and extend USWDS

Now that I've added the component, we don't need to run the gulp process — but if you ever need to modify the banner, you'll need to `npm install` and run `gulp` to generate the css.


thisisdano:
Fixes #129

thisisdano:
This adds a number of USWDS assets and fonts to the repo, but our site doesn't reference the fonts, so they don't add to the footprint.

PR #196, opened 2017-11-07
Copy change to the U.S. Digital Registry page

jeremyzilar:


jeremyzilar:
Here is the preview: https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/us-digital-registry-hotfix/services/u-s-digital-registry/

ToniBonittoGSA:
Am getting an odd string on the end of the URL; link should just be: https://usdigitalregistry.digitalgov.gov/

jeremyzilar:
@ToniBonittoGSA I think that might be something that HubSpot is entering on all `digitalgov.gov` urls. :/ We can fix or look a that separately.

ToniBonittoGSA:
Hmm.. it's also on the Register button, but not the MAX register link in last paragraph

All 3 links are fine on the live site https://www.digitalgov.gov/services/u-s-digital-registry/

jeremyzilar:
I have a feeling that this query string might only appear on domains that are not `digitalgov.gov`. This is not on DEMO or the live site

PR #199, opened 2017-11-07
Updating the markdown for list items

jeremyzilar:
The last grouping of list items was not rendering in the markdown.

I changed the dashes to asterisks in the list items, and added a carriage return to the text above. That seemed to work.

Here is the preview: https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/hotfix-federalist-event/event/workshop-making-government-websites-with-github-federalist-us-web-design-standards/


thisisdano:
Here's an example of a release note item:

- Fixed a problem with the way a list displayed in the [Making Government Websites with GitHub, Federalist, & U.S. Web Design Standards](#) event.

thisisdano:
We shouldn't use `, &` in titles. Either write out `and` or remove the comma.

PR #261, opened 2017-12-08
Changing branch new files are created in

jeremyzilar:
When creating new files in the new file generator, we should be creating them in `MASTER`.

`https://github.com/GSA/digitalgov.gov/new/master/content/events...`

They are currently pointing to DEMO

`https://github.com/GSA/digitalgov.gov/new/demo/content/events...`

This change sets the new file generator to create files in master branch.

https://www.digitalgov.gov/new-post/

### DIFF
--- a/content/pages/new-doc.html
+++ b/content/pages/new-doc.html
@@ -38,7 +38,7 @@ css: [front-matter.css]
   <div class="gcol gcol2">
     <pre id="filename" class="mttr">***</pre>
     <pre id="post-matter" class="mttr">---</pre>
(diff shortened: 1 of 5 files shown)


thisisdano:
I'm noticing that the post/doc/event links don't work on preview branches since they aren't absURL'd — pushing that change and I'll re-check.

thisisdano:
Also, getting the following error when I build:
```
ERROR 2017/12/11 10:08:01 Error while rendering "page": template: /path/to/digitalgov.gov/themes/digital.gov/layouts/_default/single.html:10:9: executing "content" at <partial "breadcrumbs...>: error calling partial: template: theme/partials/breadcrumbs.html:10:20: executing "breadcrumb" at <.currentPage.Parent>: can't evaluate field Parent in type interface {}
```

We'll see if it builds in Federalist...

thisisdano:
Well it built OK in Federalist, so I don't know what the deal is with that error. I think it's all good to go tho??

PR #275, opened 2017-12-18
UX/Resource audit edits

ToniBonittoGSA:
https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/ToniBonittoGSA-patch-5/resources/digitalgov-user-experience-resources/ 

Done:
- Revised summary
- Removed H3 tags on numbered list
- Removed H5 hashes at end

Later: 
* Unsure how to align image to right of first line of text.

Can we do something like these: https://steemit.com/howto/@tajo/the-ultimate-guide-to-image-alignment-and-links-in-markdown-become-a-pro-right-now-with-these-easy-tips ?
 ~ code like in #3 is the one that would be used here.


ToniBonittoGSA:
updated orig. comment; added link

PR #279, opened 2017-12-18
Fixing the mobile breakpoint css

jeremyzilar:
There is a bug that is making the mobile view of resource pages wider than it should be on mobile viewports. This CSS fix corrects for that.

Here is a preview: https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/hotfix-mobile-breakpoint/resources/
_(Make sure you view it on your phone 📱 )_


ToniBonittoGSA:
The wrap fix looks good; just need to remove gap at the **Digital Strategy** section

> * Digital Government Strategy (May 2012)
> 
>   * Agency Digital Strategy Pages

(visible in screen caps below)


# **Before**

before mobile view fix resources
---
# **After**

after mobile view resource fixed
---

Issue #260, opened 2017-12-08
Add relevant resources to Multilingual Community page


jeremyzilar:
In the previous version of the site, we had links on the page to the best Mutlingual resources within DigitalGov.

Here is the old version: https://web.archive.org/web/20170704155625/https://www.digitalgov.gov/communities/government-multilingual-websites-community/

They did not look so great in the old version, but they were there. Now they are not.

We can easily add them back, but they will be a list of links to resources — not a feed from the `multilingual` tag page like before.

By making it a list of resources, we (or the community manager) will be able to hand-pick the resources/blog-posts/tools that are the most valuable for the community.

---

Here are the links that were there.

### Getting Started
- [Top 10 Best Practices for Multilingual Websites](https://www.digitalgov.gov/resources/top-10-best-practices-for-multilingual-websites/)
- [Agencias y Departamentos del Gobierno Federal](https://gobierno.usa.gov/agencias-federales/a)
- [Spanish Language Style Guide and Glossaries](https://www.digitalgov.gov/resources/spanish-language-style-guide-and-glossaries/)
- [Accessibility Editorial Guidelines for Videos](https://www.digitalgov.gov/2012/08/01/accessibility-editorial-guidelines-for-youtube/)
- [Lost in Translation](https://www.digitalgov.gov/2012/10/01/automated-translation-good-solution-or-not/)

### Featured Articles
- [A Guide to Publishing Multilingual Digital Content](https://www.digitalgov.gov/2014/07/01/multilingual-digital-content/)
- [Multiple Languages, Single Platforms: the World Digital Library](https://www.digitalgov.gov/2015/04/09/multiple-languages-single-platforms-the-world-digital-library/)
- [Trends on Tuesday: Hispanic Millennials Provide Insight on Mobile Future](https://www.digitalgov.gov/2014/05/27/trends-on-tuesday-hispanic-millennials-provide-insight-on-mobile-future/)
- [Persona Development Case Study: NCI and Spanish Language Outreach](https://www.digitalgov.gov/2015/03/02/persona-development-case-study-nci-and-spanish-language-outreach/)
- [More on Multilingual »](https://www.digitalgov.gov/tag/multilingual/)

jeremyzilar:
The next step is to have Laura look over the links above and see if there are any other tools, resources, links, articles that she would like to link to. They don't have to be links to things on digitalgov.gov. 

Any links that are there, should the most valuable links for people in the multilingual community.

[...]

PR #313, opened 2018-01-09
Added in Multilingual resources to the community page

jeremyzilar:
A number of the related resources got dropped from the Multilingual Page migration. This fix brings those back. Here is the issue: https://github.com/GSA/digitalgov.gov/issues/260


ToniBonittoGSA:
@jeremyzilar I clicked on the Details link to try to view but get this error msg: 

screen shot 2018-01-09 at 5 25 22 pm

Issue #336, opened 2018-01-19
Event: February Plain Language Monthly Meeting


SoncerrayBolling:
# The February Meeting of PLAIN

**Date:** Tuesday, Jan 10, 2018, 2-3pm ET
**Summary** 
**Description** _(min ~200 words)_ 

We Heart Our Volunteers!

If you're currently volunteering for PLAIN, whether training, posting social media, or anything else, please come. 

If you're interested in volunteering for PLAIN, please come. This meeting will be focused in person, but we'll have a conference bridge line so people who are remote can still join in the discussion.

One thing: To volunteer, you have to be a federal employee. Sorry. We know many of you are eager to help us out, but according to our charter rules, we can only use feds.
Be prepared to discuss what you'd like to do. Here are some ideas:

* social media support
* community management (posting articles to the list serv, prompting conversations)
* helping update website using Github
* monthly meeting coordinator (topics, space, speakers)
And more!

---
**Tasks:**
- [ ] 30min meeting w/ presenters
- [x] Send calendar invite once event date has been determined
- [x] locate and reserve event room
- [x] Create event in Eventbrite https://www.eventbrite.com/e/february-plain-monthly-meeting-registration-42365715997
- [x] Post event page on DigitalGov
- [x] Schedule emails through Eventbrite
- [x] Schedule meeting with SME to review event logistics
- [ ] Work with DG team to promote via Social Media platforms
- [x] Track registration and provide event partner weekly registration updates
- [x] Set up evaluation in Google Forms and share with POC https://docs.google.com/forms/d/1wqlscR6xmh5yVmymADJh-1QtGGaiivwCG6Al51eZquY/edit?usp=sharing
- [ ] Update metrics spreadsheet
- [ ] Schedule post event/training retro
- [ ] Send post-event email with evaluation and any additional resources
- [ ] Share evaluation results with event partner and presenters
- [ ] Conduct retro

PR #337, opened 2018-01-19
Add new event: February PLAIN Meeting

SoncerrayBolling:
**Here is the issue:** https://github.com/GSA/digitalgov.gov/issues/336

**Preview:** https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/SoncerrayBolling-patch-1/


jeremyzilar:
For some reason, this is not building. I am looking into it. 👀

Issue #340, opened 2018-01-23
U.S. Web Design Standards Monthly Users Call


AndreaSigz:
# U.S. Web Design Standards Monthly Users Call

**Date:** Friday, January 26, 2018, 1-1:30pm ET
**Summary** _(one sentence)_ — In our next monthly users call, the U.S. Web Design Standards core team will share what they’re working on in 2018.
**Description** _(min ~200 words)_ — 

In our next monthly users call, [the U.S. Web Design Standards](https://standards.usa.gov/) core team will share what they’re working on in 2018, including:

- changing our name to more accurately describe what we are and what we do
- more specific design and implementation guidance
- easier ways to prototype and build consistently and incrementally
- clearer connections between user research and guidance
- a path to contribute components and research back to the system

Please join us on Zoom, on **January 26 at 1pm ET**, to hear from the team and ask your questions. Register here: https://www.eventbrite.com/e/january-us-web-design-standards-users-call-tickets-42250600684

Presenters:

- Maya Benari
- John Donmoyer
- Dan Williams

_This talk is open to federal employees who work on digital projects that could use the Web Design Standards._

---

- [ ] 30min meeting w/ presenters
- [ ] Send calendar invite once event date has been determined
- [ ] Request video title card
- [ ] Create event in Eventbrite
- [ ] Create event in YouTube live
- [ ] Post event page on DigitalGov
- [ ] Schedule emails through Eventbrite
- [ ] Schedule real time captioning through FedRelay
- [ ] Identify moderator for live chat/online questions and send moderator script
- [ ] Schedule tech check with presenters to review online platform and logistics of event
- [ ] Work with DG team to promote via Social Media platforms
- [ ] Track registration and provide event partner weekly registration updates
- [ ] Set up evaluation in Google Forms and share with POC
- [ ] Update metrics spreadsheet
- [ ] Schedule post event/training retro
- [ ] Add title card immediately following event
- [ ] Send post-event email with link to video, evaluation, and any additional resources
- [ ] Share evaluation results with event partner and presenters
- [ ] Conduct retro

jeremyzilar:
Will this talk be over ZOOM, or YouTube LIVE?
If over ZOOM, it might be good to state that.

jeremyzilar:
I updated a few things in the desc, like adding a link to the Standards and moving the presenters up.

AndreaSigz:
Just added the phrase 'on Zoom' to the description. Thx!

PR #343, opened 2018-01-24
Add new event: USWDS

jeremyzilar:
Added in the new USWDS event: https://github.com/GSA/digitalgov.gov/issues/340

**Preview:**  https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/uswds-event/event/us-web-design-standards-monthly-users-call/


ToniBonittoGSA:
@jeremyzilar do you need to change Standards to System yet?

Issue #345, opened 2018-01-25
Incorrect author in FB OG meta code


ToniBonittoGSA:
I was going to share the new Events page https://www.digitalgov.gov/events/ to Facebook but noticed that title and summary are wrong. 

When I looked at the code, I saw that Will Sullivan is listed as author in the head for FB OG (this is site-wide)

**Q**: 
Can JZ/Dan do a mass find & replace to change: 

&lt;meta property="article:author" content="willsullivan" /&gt;

to: 

&lt;meta property="article:author" content="" /&gt; 

so the correct author profile URL(s) or ID(s) can be specified when known?

jeremyzilar:
Yeah, I can work on this now. Thank you for flagging @ToniBonittoGSA

jeremyzilar:
I'd also like to remove the `OG:image` values and let Twitter/Facebook/Slack pick up what ever images they want to in the post.

jeremyzilar:
Reading up on:
https://developers.facebook.com/docs/reference/opengraph/object-type/article/
http://ogp.me/

PR #347, opened 2018-01-25
Fixed Open Graph meta tags

jeremyzilar:
A fix to the data in the OG meta tags in the head of the page to correct this **issue:** https://github.com/GSA/digitalgov.gov/issues/345

## What was made better
- removed the `article:author` property all together, since that property only accepts single authors and we often publish with multiple authors.
- removed the `last-modified` property because that is hard to access as the moment, and non-essential
- made the featured image conditional

**Preview:** https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/og-image-card/

### DIFF
--- a/themes/digital.gov/layouts/partials/head.html
+++ b/themes/digital.gov/layouts/partials/head.html
@@ -53,6 +53,7 @@
   <meta property="og:type" content="{{ $.Scratch.Get "pagetype" }}">
   <meta property="og:url" content="{{ .Permalink }}" />
   <meta property="og:site_name" content="DigitalGov">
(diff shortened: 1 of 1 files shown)


ToniBonittoGSA:
@jeremyzilar Looks like you can have more than one author; see OG Properties on 
https://developers.facebook.com/docs/reference/opengraph/object-type/article/ 

Do you want to put back in?

jeremyzilar:
@ToniBonittoGSA OH! Look at that. 
Oh wait, they need to be Facebook IDs of the authors 😞 


We don't really need any of the items that starts with `article:somthing` 
Here is an overview: https://moz.com/blog/meta-data-templates-123

jeremyzilar:
This will also only show an image, when there is an featured image in the front matter

PR #421, opened 2018-02-12
Spanish Lang. Style Guide pg: minor corrections

ToniBonittoGSA:
Done:
 - Changed Guide in title and copy and verb tenses to plural (singular 'Guide' was causing confusion; users expected file to download)
 - updated summary

To-do:
 - update url
 - add alias
 - add corrections from Laura's Google Doc; will put in Slack


ToniBonittoGSA:
added 3rd to-do

PR #457, opened 2018-02-23
Add new event: A Deep Dive Into Top Tasks with Gerry McGovern

jthalls:
**A Deep Dive Into Top Tasks with Gerry McGovern** 
This talk will focus on how using the top tasks approach will help you deliver what your customers want to make it easier and faster for people to do things they need to do with government


AndreaSigz:
I am merging this PR

Issue #454, opened 2018-02-22
Event: Top Tasks Refresh with Gerry McGovern


AndreaSigz:
# Top Tasks Refresh with Gerry McGovern

Live page: https://www.digitalgov.gov/event/a-deep-dive-into-top-tasks-with-gerry-mcgovern/

**Date:** Wednesday, April 11, 2018, 11am-12pm ET
**Summary** _(one sentence)_ — This talk will focus on how using the top tasks approach will help you deliver what your customers want to make it easier and faster for people to do things they need to do with government.
**Description** _(min ~200 words)_ — 
 Google doc with description: https://docs.google.com/document/d/1lEVVkGw3QFK70QASh4gb9oXiMTSd2zPAPvxMBlnUBCM/edit#

---

- [x] 30min meeting w/ presenters
- [x] Send calendar invite once event date has been determined
- [x] Request video title card
- [x] Create event in Eventbrite
- [x] Create event in YouTube live
- [x] Post event page on DigitalGov
- [ ] Schedule emails through Eventbrite
- [x] Schedule real time captioning through FedRelay http://www.fedrcc.us//Enter.aspx?EventID=3608989&CustomerID=321
- [x] Identify moderator for live chat/online questions and send moderator script
- [x] Schedule tech check with presenters to review online platform and logistics of event
- [x] Work with DG team to promote via Social Media platforms
- [x] Track registration and provide event partner weekly registration updates
- [x] Set up evaluation in Google Forms and share with POC
- [x] Update metrics spreadsheet
- [x] Schedule post event/training retro
- [x] Add title card immediately following event
- [x] Send post-event email with link to video, evaluation, and any additional resources
- [x] Share evaluation results with event partner and presenters
- [x] Conduct retro

AndreaSigz:
Let me know next steps. Thanks!

jeremyzilar:
Can you add the event description to the top of the issue?

AndreaSigz:
Janelle asked me to link to it, but maybe I misunderstood

jeremyzilar:
@AndreaSigz can we update the summary to:

`summary:` This one-hour talk will help you deliver what your customers want and defend your decisions based on clear evidence, by learning about the core elements of Top Tasks.

jeremyzilar:
@jthalls @SoncerrayBolling can you get the `YouTube ID` and `Registration URL`?

jeremyzilar:
[...]

PR #462, opened 2018-02-27
Added image to Gerry McGovern event page

jeremyzilar:
### Details of the event: https://github.com/GSA/digitalgov.gov/issues/454

**Event preview:** https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/Gerry/event/a-deep-dive-into-top-tasks-with-gerry-mcgovern/


jeremyzilar:
Looks like this event is already live on the site. https://www.digitalgov.gov/event/a-deep-dive-into-top-tasks-with-gerry-mcgovern/

PR #518, opened 2018-03-21
Fixing chat box embed

jeremyzilar:
### Fixes
There was an issue with the chat box where the chat box embed was not working on event pages.
This fixes that error, I hope.

**Preview:** https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/YouTube-Chat-embed-fix/event/conversations-on-public-impact-how-data-trusts-can-preserve-protect-your-digital-initiatives/

### DIFF
--- a/themes/digital.gov/layouts/partials/event-stage.html
+++ b/themes/digital.gov/layouts/partials/event-stage.html
@@ -3,7 +3,7 @@
     <div class="youtube-player" data-id="{{ .Params.youtube_id }}"></div>
   </div>
   <div class="chat-box">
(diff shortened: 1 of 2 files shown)


jeremyzilar:
@jthalls The code for this is in and _should_ make the chat box appear once it goes live

jeremyzilar:
Ok, so I noticed that the video stage was not appearing on the event page for today. There are some error with the way that HUGO is putting out dates and comparing those dates with the in-browser dates.

I am going to remove the logic that hides the videos for now, as there is an event in an hour.

PR #519, opened 2018-03-21
Updating Design System info

jeremyzilar:
## What we made better
- updated design system URL
- updated core team roster

**Preview:** https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/about-fix/about/


ToniBonittoGSA:
@jeremyzilar is this the only page that was updated https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/about-fix/about/ ?

jeremyzilar:
@ToniBonittoGSA yes, this was the only page that was updated

PR #622, opened 2018-05-30
Fixing time EST

jeremyzilar:
## Fixed 

- Event pages
- Event listings
- feedback boxes

**Preview:** https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/time-est/


jeremyzilar:
No need to fix. We are going to stick with ET

Issue #717, opened 2018-07-24
All "posts" page on digital.gov


jeremyzilar:
At the moment, we don't have an "All Posts" page on digital.gov. The only place to view our blog posts, is on the homepage, and we are only surfacing the last few.

## it should include …
- [x] The last 20 blog posts
- [x] navigation to previous 20 blog posts ( and so on ... )

## it's done when …
- [x] this is live and working at digital.gov/posts/

jeremyzilar:
This has been completed 🎉👾

PR #719, opened 2018-07-25
All posts page

jeremyzilar:
## Fixing https://github.com/GSA/digitalgov.gov/issues/717

This fix is aimed at creating a means of listing out, or making visible, all the blog posts on digital.gov, in a reverse chronological order.

While we can navigate digital.gov by tag, category, author, or through search, we don't have a way to scroll back through the blog posts over time. The only place to view our blog posts, is on the homepage, and we are only surfacing the last few.

**Preview:** https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/all-posts-page/posts/

_Note: There are 30 posts per page._

image


jeremyzilar:
Notes from Gail @gailswanson — 
- this page is all about making it easier for people to scan. Optimize for that.
- loose the pagination at the top
- less padding, more density
- reduce the thumbnail to smaller image (75px x 75)
- reduce to 15 results per page
- include a means to navigate by tag, categories, author at the top or in the right rail
- use words with next and previous links in the pagination
- remove the jump to last/first links in the pagination
- change heading to link to first page

thisisdano:
I would consider not organizing this via brute force pagination — that is dozens of pages of reverse-chron. Rather, I would organize it by date, so one could go back to a specific year/month, and list out those years/months in a sidebar

thisisdano:
I'd be curious to know why 15 is preferable to 25/30. It feels like too few.

Issue #723, opened 2018-07-31
Load Scripts through Google Tag Manager


jeremyzilar:
Currently, we are including our scripts on digital.gov directly on the page. This results in a lot of overhead processing on the part of the browser and additional HTTPS calls.

I would like to start by moving our Google Analytics scripts to Google Tag Manager as a first step.

## it should include …
- [ ] All GA tags should be loaded by Tag Manager
- [ ] A document in the wiki that explains what is being included via Tag Manager, and how to update it.

## it's done when …
- [ ] it is seamlessly working

jeremyzilar:
Thank you @laurenancona !
DONE — I added the GTM tags to the page, both of them.

I didn't remove the GA tags just yet.

**Preview:** https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/tag-manager/

laurenancona:
Scripts to be migrated:
- [ ] **Google Analytics pageview tag: UA-47271808-1**
```
<!-- Google Analytics — Global Site Tag (gtag.js) \| because our site has been around a lot longer than DAP -->
  <script async src="https://www.googletagmanager.com/gtag/js?id=UA-47271808-1"></script>
  <script>
  window.dataLayer = window.dataLayer \|\| [];
  function gtag(){dataLayer.push(arguments)};
  gtag('js', new Date());
   
  gtag('config', 'UA-47271808-1');
  </script>
```
_This and any additional custom `gtag` config should be migrated to GTM - do you know of any other event or other custom tracking deployed on the site?_


- [ ] **Custom HTML tag: DAP (UA-33523145-1)**
```
<!-- DAP — Digital Analytics Program \| Learn more: https://digital.gov/dap/ -->
 <script language="javascript" id="_fed_an_ua_tag" src="https://dap.digitalgov.gov/Universal-Federated-Analytics-Min.js?ver=true&agency=GSA"></script>
```

jeremyzilar:
OK, I removed DAP and GA from our pages. 
**Preview:** https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/tag-manager/

Also worth taking a look at how we implemented the Video Tracking code in DAP https://github.com/GSA/digitalgov.gov/issues/561

jeremyzilar:
This has been completed — thanks to @laurenancona  🎉👾

PR #724, opened 2018-08-01
Implement Google tag manager on digital.gov

jeremyzilar:
## Fixes https://github.com/GSA/digitalgov.gov/issues/723

**Preview:** https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/tag-manager/


laurenancona:
@jeremyzilar [These are the lines ](https://github.com/GSA/digitalgov.gov/blob/3f9c8bd0ed3e0c0f7bf4f61104189ad03c763e4d/themes/digital.gov/layouts/partials/head.html#L148-L159) to remove (in a 2nd PR?) when we're ready to swap those analytics tags out of source and into GTM

PR #730, opened 2018-08-09
New DAP page

ToniBonittoGSA:
From Tim's doc

Preview: https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/TB-DAP-Certified-Analyst/services/dap/certified-analyst/ 

Needs: 
linked button
featured image


ToniBonittoGSA:
@jeremyzilar 
I found the Ogilvy DAP logos; there are 2 versions: 

1 - 

digitalgov_dap2


2 - 

digitalgov_stacked dap3

tdlowden:
One comment on the PDFs-- they aren't actually a 1.4 or 1.7 MB download of a PDF right from the link. I hosted them on Github so that people could view without an actual download. So, if they choose to download from Github, that's one thing, but I feel like the `(PDF, 1.7MB)` language makes it seem as if they are downloading right from the link, which isn't the case here.

PR #746, opened 2018-08-20
Update certified-analyst.md

tdlowden:
also made a few content adjustments


jeremyzilar:
Here is a preview: https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/certified-analyst-changes/services/dap/certified-analyst/

ToniBonittoGSA:
@jeremyzilar @tdlowden   if we set the subject line in the email link code, can we shorten from

> To request the link to take the exam, email us at dap@support.digitalgov.gov with the subject: “DAP Certification Exam Request”.

to just

> To request the link to take the exam, email us at dap@support.digitalgov.gov.

tdlowden:
That's assuming people click on the `mailto` link (something I personally never do bc I use browser gmail rather than desktop mail client). So I think adding it in the link is a great idea, but I would prefer the language about subject line stay there as well.

PR #772, opened 2018-09-12
adding caching to partials

jeremyzilar:
## What we made better

After running a few scripts to assess the performance of HUGO, 
https://gohugo.io/troubleshooting/build-performance/

I was able to determine that a number of the partials and sub-files we have can be cached.

See: https://gohugo.io/functions/partialcached/

This should speed up the build of the site.

Preview: https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/partial-caching/


jeremyzilar:
Actually, this is breaking a few things. I need to look at it further

jeremyzilar:
Not needed. This work has been rolled into the redesign

PR #828, opened 2018-11-06
Fix typo in 11/6 FedRAMP event description

lcrabb419:
Replaced "CSPs" with "Agency partners" in the following sentence: "the resources available to help CSPs through the FedRAMP authorization process."


jeremyzilar:
Closing. No need to fix.

PR #877, opened 2018-12-18
Add new post: Looking Back at the PIF

jeremyzilar:
**Preview:** https://federalist-proxy-staging.app.cloud.gov/preview/gsa/digitalgov.gov/pifs-2018-in-review/2018/12/19/looking-back-at-pifs-in-2018/


jeremyzilar:
OK @ToniBonittoGSA This is the PIF post that will go up tomorrow morning after the GSA publishes their version.

We will need to update the URL in the `canonical: URL-GOES-HERE` area and in the top of the post.
The homepage should also be set.

**There are a few additional changes going up with this post.**
- some NPM fixes
- new author files for the new authors...
- and the inclusion of the DECK field in Blog posts (optional)

PR #891, opened 2019-02-06
Certificates shutdown

jeremyzilar:
I would like to get this guidance out in-front of the people in govt who are actively preparing for the next shutdown.

The need for this guidance comes directly out of conversations we've had with people in the Web Council who are needing to find the best way to prevent security certificates from expiring during a shutdown period.

This is not guidance about managing domains and certificates long-term, but rather it's goal is to solve a number of impending problems for agencies in the near-term.

If this information got there tomorrow, agencies would have a significantly good chance of averting these types of errors before the next shutdown happens.

Preview: https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/cerificates-shutdown/resources/how-to-prevent-security-certificates-from-expiring-during-shutdown/


ToniBonittoGSA:
@jeremyzilar fix title case to:
How to Prevent Security Certificates From Expiring During a Lapse in Operations

jeremyzilar:
Fixed @ToniBonittoGSA 👍

PR #945, opened 2019-03-21
event page for api COP event

gbinal:


gbinal:
Looks good - thanks, @Schultea.  

Could I go ahead and get a merge to make the page live?  It's ready for circulation and then I'm going to update it with some links later today but nothing that the page needs to wait on.

PR #1102, opened 2019-06-18
A banner about DEMO for old site

jeremyzilar:
This introduces a banner for the site that appears at the top of each page on the old digital.gov.

image

This is to be launched when we send out the link to the DEMO site


jeremyzilar:
We're not going to work on this at this time. 
We either don't have the bandwidth, or the idea/issue is no longer relevant.

PR #1107, opened 2019-06-18
update to open innovation summit page

twgsa16:
Suggesting an update for Open Innovation Summit event page. Need to confirm that we'll point directly to PDF hosted on White House site and separately provide presentations (PPT converted to PDF), but in the meantime, providing for review and to start staging.


jeremyzilar:
@twgsa16 let us know when you have the link to the PDF, and we'll get this up.
or we can remove that additional link, get up what you have here, and then add the link back in once you have it? + @ToniBonittoGSA

twgsa16:
PDF file is too large for attachment. Can be downloaded via Google Drive: https://drive.google.com/drive/folders/1XsIQYV6oy5IX5X6IC9lEinD7awZ2cvfM

jeremyzilar:
@twgsa16 I am moving this to a new branch to able to add the PDFs

PR #1134, opened 2019-07-01
change of guests in copy

Schultea:


ToniBonittoGSA:
this was same edit in https://github.com/GSA/digitalgov.gov/pull/1137 
closing.

PR #1181, opened 2019-07-23
Removing CC link

EssbieWGT:
Just taking out the CC link now that the event has past.


ToniBonittoGSA:
This was done in another PR; closing

PR #1215, opened 2019-08-08
Event: Plain Summit 2019

jeremyzilar:
### A day of talks and conversations on how to advance plain language in government communications

**Preview:** https://cg-06ab120d-836f-49a2-bc22-9dfb1585c3c6.app.cloud.gov/preview/gsa/digitalgov.gov/plain-summit-2019/event/2019/09/05/plain-language-summit-2019/


jeremyzilar:
@lcrabb419 — There are two things that we might need to add in here
- Correct room number
- Some text about how we'll be having people join this online

lcrabb419:
> @lcrabb419 — There are two things that we might need to add in here
> 
> * Correct room number
> * Some text about how we'll be having people join this online

- FYI, the room number already was correct

PR #1258, opened 2019-08-29
Fixing flash on USWDS banner

jeremyzilar:
From @eddietejeda —

> "Hey @thisisdano @jeremyzilar , since upgrading to 2.0, I noticed that the top government banner flickers on page reloads. This became very apparent in the new digital.gov site, the pages are a bit bigger than in our demo sites. Is there a solution to this?"

## What we made better

This change adds `hidden` to the element that contains the additional "Here's how you know" content in the USWDS Gov banner.

```
<div class="usa-banner__content usa-accordion__content" id="gov-banner" hidden>
```

This makes it so that the element is hidden by default when the page loads. After the USWDS JS loads, it still applies `hidden` to this element, but since it is already hidden, there is no visible effect and no flash

### DIFF
--- a/themes/digital.gov/layouts/partials/usa-banner.html
+++ b/themes/digital.gov/layouts/partials/usa-banner.html
@@ -20,7 +20,7 @@
         </button>
       </div>
     </header>
-    <div class="usa-banner__content usa-accordion__content" id="gov-banner">
+    <div class="usa-banner__content usa-accordion__content" id="gov-banner" hidden>
       <div class="grid-row grid-gap-lg">
         <div class="usa-banner__guidance tablet:grid-col-6">
           {{- $uswds_asset := "icon-dot-gov.svg" -}}


thisisdano:
So, afaik this should be not be a new behavior in USWDS 2. I’ll need to dig into the history, but even in USWDS 1.x we should be activating *hidden* only on the JavaScript load. This is to assure that all necessary content is visible even if the JavaScript does not load. 

This solution will work, but at the expense of the intention of the no-js design. It’s fair to question whether the banner content needs to be subject to our visible-even-with-no-js

thisisdano:
Sorry, closed this PR unintentionally!

jeremyzilar:
@thisisdano In the spirit of being a good example for others, I am happy to hold on this till you dig further. And maybe in the meantime, I can compile and minify our 22 `.js` files and remove any unused `.js` files to make the overall size smaller.

thisisdano:
A bit more context: https://github.com/uswds/uswds/pull/3075

jeremyzilar:
This is not the right approach with the USWDS, so I am closing for now. Happy to re-open if we decide to do this at a later point

PR #1288, opened 2019-09-23
Update _index.md

RachelF:
Clarify scope of fed web policies; requirements cover exec branch only; optional for other fed .gov sites


jeremyzilar:
FYI @tdlowden

PR #1316, opened 2019-10-01
Updating the header on DEMO

jeremyzilar:
We've made a considerable amount of infrastructure changes to Digital.gov, and we are aiming to get them launched. In order to do that, we are making some changes to DEMO to make it look a little more like the current site, in the hopes that it will make the leap to a new site feel like like a leap.

**Preview:** https://cg-06ab120d-836f-49a2-bc22-9dfb1585c3c6.app.cloud.gov/preview/gsa/digitalgov.gov/new-demo-header/

This change merges into DEMO


jeremyzilar:
Need to fix:

- [ ] broken images in gov banner
- [ ] broken images in mobile nav

PR #1409, opened 2019-10-24
USWDS: Reduced our overall CSS filesize by 86.58%

jeremyzilar:
A short explanation of how we started out prototyping with the [USWDS](https://designsystem.digital.gov/) and paired it down to just the styles we needed and **cut our overall CSS filesize by 86.58%** 🎉

---

We've been aiming to build the new Digital.gov completely using the design tokens found in USWDS 2.0. 

## Expressing with Design Tokens
The [design tokens](https://designsystem.digital.gov/design-tokens/) are like the legos of the USWDS — each is a building block for reliably making your website. 

> This degree of choice can slow down design work and make communication between designer and developer unnecessarily granular. USWDS seeks to maximize design efficiency and improve communication with design tokens: the discrete palettes of values from which we base all our visual design.

> Design tokens are a limited set of discrete options, just like a scale of musical notes is drawn from the spectrum of all possible frequencies. Or like the presets on a car radio — not every option, just a specific selection.

You can implement design tokens through the utility classes (via HTML classnames) or via mixins in your scss files (these get rendered into CSS).

#### Design Tokens via Utility Classes
_See: https://designsystem.digital.gov/utilities/_
These are styles that are only applied via classnames in the HTML. Super fast and easy to apply, change, and a great way to get a sense for the design language they offer. Perfect for prototyping!
```
    <div id="event" class="margin-3 padding-2 width-15">
      ...
    </div>
```

#### Design Tokens via Mixins
The mixins do the same thing, they apply the design tokens, but they offer greater control over and are a little easier to use if you are familiar with CSS.
```
    #event{
      @include u-margin(3);
      @include u-padding(2);
      @include u-width(15);
    }
```

## Prototyping our way forward
When we started out, the USWDS utility classes provided us with the ability to quickly try out a number of different layouts in the early stages of our build.  And once we were pretty happy with a particular part of the site, we'd start moving the styles over to a SCSS file and use the mixins to refine the design. 

Once we started to become more sure of our direction, we were ready to start optimizing our layouts for performance. And one of the big flags that stood out in our Lighthouse reports was that we had _a lot_ of unused CSS on the site, mostly due to the USWDS utilities.

For the utilities to be 'available' for use, the design system needs to build out the CSS every possible utility `margin-3`, `margin-2`, `margin-1`, `margin-2px`, ... and so on. This is what makes them so flexible and easy to use directly in the HTML. But in the CSS, a lot of 'potential' css goes unused in a final design.

## Optimizing USWDS for performance
Fortunately, the USWDS settings allow us to turn off the utilities one by one, which essentially stops it from building out those blocks of CSS for each utility. And that's what we did. We actually turned off all of the utility classes (with the exception of `$order-settings` and `$display-settings` which are convenient classes to have when making layouts. Then one by one, we did a find-and-replace to make sure we didn't leave any utilities hanging around. In all, it took a few hours to ensure that everything had been moved over to SCSS files.

Doing this greatly cut down on the overall CSS filesize by 86.58%.
- **Starting size:** ~514 KB — https://demo.digital.gov/
- **Optimized size:** ~69 KB  — [Preview URL](https://cg-06ab120d-836f-49a2-bc22-9dfb1585c3c6.app.cloud.gov/preview/gsa/digitalgov.gov/remove-utilities/)

_These tests were ran against the homepage of https://demo.digital.gov only._

---

## Fixes remaining
- [x] Need to go in find the few areas in the site that were using utilities, and swap them for USWDS mixins.


thisisdano:
NICE.

PR #1455, opened 2019-11-06
updated url

DanielJPino-ux:
had an error in the url for making it "public' this has now been resolved - the correct live link is here: https://www.youtube.com/watch?v=MRfsXVOicC8&t=115s

### DIFF
--- a/content/events/2019/11/2019-11-04-foundations-agile-i.md
+++ b/content/events/2019/11/2019-11-04-foundations-agile-i.md
@@ -12,8 +12,7 @@ date: 2019-11-04 13:30:00 -0500
 end_date: 2019-11-04 14:30:00 -0500
 event_organizer: DigitalGov University
 host: 18F
-registration_url: https://www.eventbrite.com/e/agile-talks-foundations-of-agile-part-i-registration-78043727891
-youtube_id:
+youtube_id:MRfsXVOicC8&t=115s
 
 ---
 


jeremyzilar:
## How to read the YouTube URLs

The YouTube ID can be tricky to grab. It is in the URL, but you also have to watch out for the YouTube timecode, because that is often in there as well. And note, this is how every web editor at every news org on the planet does this.

### Breaking down the URL
The full URL: `https://www.youtube.com/watch?v=MRfsXVOicC8&t=115s`

That first part is just the URL to the YouTube app `https://www.youtube.com/watch`.

That is followed by the query string. These think of these as "options" that you're telling the website to have when it loads the website. 

NOTE: All query strings starts with a `?` and are followed by specific options or parameters that look like `a=applesauce` or `z=zebrasauce`. When there are multiple query strings, they get joined by an `&` symbol.

Here is the query string in the YouTube that you passed over: `?v=MRfsXVOicC8&t=115s`. You'll notice that there are two parameters:
1. the Video ID, as represented by the `v=MRfsXVOicC8`
2. the timecode, as represented by the `t=115s`. This is the optional setting for how many seconds into the video you want it to start playing when the user loads the video.

The video ID is essential for loading a video in YouTube. The timecode is optional. 

- Try it out with the timecode set to 2000 seconds! https://www.youtube.com/watch?v=MRfsXVOicC8&t=2000s
- Try it out without the timecode at all https://www.youtube.com/watch?v=MRfsXVOicC8

## How it works in Digital.gov

In event pages on Digital.gov, we typically pass in a video ID like this:

```
youtube_id: MRfsXVOicC8
```
And when the page builds, these two things get added together: 
`https://www.youtube.com/watch?v=` + `MRfsXVOicC8`. 
_(the first part of the URL is baked into event page template)_

### Adding timecode option
So if you want to add in the timecode option to a video, you just need to add the additional parameter like this (don't forget the **s** at the end):
```
youtube_id: MRfsXVOicC8&t=2000s
```

### Possible errors you might encounter
Here are a few things that could go wrong here that will cause the site to not build or not display your video.

1. You forget to put a space between the `youtube_id: ` and the video ID 😬 
```
youtube_id:MRfsXVOicC8
```
2. You included an additional colon somewhere in the ID 😬 
```
youtube_id: MRfsXVOicC8:
```

jeremyzilar:
This PR fixes it if you want to take a look: https://github.com/GSA/digitalgov.gov/pull/1454
but it also removes the timecode... Is that what you want? Or do you want the timecode in there as well?

jeremyzilar:
Closing, as it duplicates https://github.com/GSA/digitalgov.gov/pull/1454

PR #1459, opened 2019-11-07
Update 2019-10-30-federal-crowdsourcing-webinar-series-episode-6-fedr…

ayerkes:
Added youtube video, removed captioning and zoom info.


jeremyzilar:
Why remove this image? It's a good bit of color on the page.
I would remove it if there was a video present, but without that, I think we should keep the image here.

Let me know if there is something about the image that is not right. I may be missing something

jeremyzilar:
Closing as this was already done in https://github.com/GSA/digitalgov.gov/pull/1460

PR #1531, opened 2019-11-20
Elevate the GSA presence on Digital.gov

jeremyzilar:
These changes introduce more prominent GSA branding to the new Digital.gov design.

**Goal:** Make it more clear that Digital.gov is a site run by the GSA, without unintentionally signaling that this site is guidance _only_ for the GSA.

### What we changed

- [x] Updated base colors in the site to `blue` from `indigo`
- [x] Adjusted the accent colors in the site
- [x] Adjusted colors to be accessible values, using the [**USWDS color tokens**](https://designsystem.digital.gov/design-tokens/color/system-tokens/)
- [x] Prominently added the GSA logo above the fold
- [x] Indicated that this site is _"published by the U.S. General Services Administration"_ — which is the right framing for this.

**Preview URL:** https://cg-06ab120d-836f-49a2-bc22-9dfb1585c3c6.app.cloud.gov/preview/gsa/digitalgov.gov/GSA-header/


thisisdano:
I definitely like the `Published by the U.S. General Services Administration` — it's really clear and helps me understand that the publication is intended for a non-GSA audience, even if it comes _from_ GSA. I also like that it connects the publisher mark to the site mission. This lends GSA's credibility to the message, where it matters more. And the logo itself is at a comfortable and confident size. It pops, it's appropriate to the placement, and it communicates confidence at the right time in the eye's scan of the page.

PR #1719, opened 2019-12-17
Fix button shortcode

jeremyzilar:
- [x] added classes to the button shortcode
- [x] added the button shortcode to the page

```
{{< button href="http://www.gpo.gov/fdsys/pkg/PLAW-104publ13/html/PLAW-104publ13.htm" text="View the Paperwork Reduction Act of 1995" >}}
```

---

**Preview:** 
https://cg-06ab120d-836f-49a2-bc22-9dfb1585c3c6.app.cloud.gov/preview/gsa/digitalgov.gov/fix-typo-and-add-resource/resources/paperwork-reduction-act-44-u-s-c-3501-et-seq/

### DIFF
--- a/content/resources/paperwork-reduction-act-44-u-s-c-3501-et-seq.md
+++ b/content/resources/paperwork-reduction-act-44-u-s-c-3501-et-seq.md
@@ -1,8 +1,8 @@
 ---
 slug: paperwork-reduction-act-44-u-s-c-3501-et-seq
 date: 2015-12-01 1:40:45 -0400
-title: Paperwork Reduction Act (44 U.S.C. 3501 et seq.)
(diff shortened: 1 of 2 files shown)


ToniBonittoGSA:
Other buttons on the site have a colored background.. 
the original: http://web.archive.org/web/20190712152033/https://digital.gov/resources/paperwork-reduction-act-44-u-s-c-3501-et-seq/ 
Can you do something similar?

jeremyzilar:
If we are going to have buttons in the body text, they should visually stand out from the text but not be louder than other elements on the page.

The border, the text color, and the fact that it’s a block element help it to stand out from the text without being too loud.

We should probably also set some limits on how long the text should be. The text in the button on the PRA page is way too long. 

What happens on mobile when the text is really long? 

When should a button like this just be a link?

jeremyzilar:
What job is the Button on the PRA page doing? 

Is it the same purpose as the button on the Challenge.gov Services page? 

If so, are they serving the same function — to clearly link to the primary source that the page is about? 

Are there cases where there might be a page that needs to prominently link to more than one source?

ToniBonittoGSA:
All very good questions!

_Note_: all of the resource pages like this have that style of button

PR #1950, opened 2020-02-03
Create CONTRIBUTING.md

saracope:
Adds CONTRIBUTING.md file


saracope:
This info would only be for contributing to the open source repo here on GitHub. I think that link would be helpful for people. So adding something in like "Interested in contributing the content on digital.gov? Check this out..." does make sense.

PR #1986, opened 2020-02-12
Fixing topics API

jeremyzilar:
This fixes the Topics API, which seemed to be broken.
I am not sure how this particular page got these two topics applied, but they were not the right "slugs", and thus caused some issues with the way it builds the JSON API.

FYI — the only way we're able to tell if an API path is broken, is by trying to use it on the workflow site.

---

**Preview:** This path `/topics/v1/json/` should be valid JSON in this preview

https://cg-06ab120d-836f-49a2-bc22-9dfb1585c3c6.app.cloud.gov/preview/gsa/digitalgov.gov/fixing-topics-API/topics/v1/json/


saracope:
Can you share the process/path you went down to figure out it was this page that was causing the issue?

PR #2025, opened 2020-02-18
Create action to close stale issues

saracope:
It looks like we can use GitHub Actions now. Creating a test workflow. 

This change will set up an Action that runs each Friday at midnight to look for stale Issues and Pull Requests.   

It will:
1. Review all Issues (includes PR's) and evaluate if they are stale. A Stale Issue is one that hasn't had any activity in 60 days. 
1. If so, adds a `[status] stale` label. 
1. Ignore any issue/pr with a label of `---- HOLD ----`.
1. Issues with the `[status] stale` label that haven't had additional activity will be automatically closed after 7 days.

Additional documentation for the [Stale action](https://github.com/actions/stale)

### DIFF
--- a/.github/workflows/close-stale-issue.yml
+++ b/.github/workflows/close-stale-issue.yml
@@ -3,7 +3,7 @@ name: "Mark stale issues and pull requests"
 # Run every Friday at midnight
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


jeremyzilar:
This is wild 🎉

PR #2042, opened 2020-02-21
Update service_codegov.md

jcastle:
Updated Code.gov service summary and topics.

### DIFF
--- a/content/services/service_codegov.md
+++ b/content/services/service_codegov.md
@@ -23,8 +23,9 @@ weight: 2
 
 # Topics that best describe this product or service
 topics:
-  - federal source code
-  - open source software
+  - open-source
+  - code
+  - data
   - api
 
 ---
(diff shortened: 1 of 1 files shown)


content/services/service_codegov.md
```diff
@@ -23,9 +23,8 @@ weight: 2
 
 # Topics that best describe this product or service
 topics:
-  - open-source
-  - code
```
saracope: I think we should keep the topics the same. They directly affect how they are displayed in the [resource](http://localhost:1313/topics/) listing. 

The `code` topic gets a more prominent page:
https://digital.gov/topics/code/

Same with `open-source`:
http://localhost:1313/topics/open-source/

While the lesser-used topics (like federal source code and open source software) would get a basic page like:
https://digital.gov/topics/fitara/

jcastle: I don't like your topics. They do adequately describe Code.gov.

Want me to create an issue to update you topics? ;-)

PR #2087, opened 2020-03-02
New authors: John Hamilton

DanielJPino-ux:

---

**Preview:**


jeremyzilar:
@DanielJPino-ux — the agency name and agency acronym can only contain one. We cannot string together multiple acronymns. I removed the fedramp one, since that is clearly included in the bio.

PR #2153, opened 2020-03-05
Rename / organize topics-list files

jeremyzilar:
This abstracts out the styles for just the topics list (those topics that can be found on the article and event pages). They were nested in with other styles.

Now there is a corresponding partial and scss file and classname, all with the `topics-list` name.

---

**Preview:** https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.app.cloud.gov/preview/gsa/digitalgov.gov/rename-topics-list-files/2020/02/27/call-for-government-ux-summit/


jeremyzilar:
@saracope This is all fixed!

PR #2165, opened 2020-03-06
Add preview link action

saracope:
Changes include:
* add a github action to post the Federalist preview link to PR's

The workflow only runs on pull requests to the master branch.
It will automatically add a comment to the PR with a link to view the Federalist preview.
The link is generated based on the branch name.

The next todo would be to post the comment only after the federalist branch is done building. Right now it doesn't take that into consideration. This could cause some initial confusion with users BUT I need to see how it interacts with the current PR checks in order to add that piece.

### DIFF
--- a/.github/workflows/pr-preview-link.yml
+++ b/.github/workflows/pr-preview-link.yml
@@ -1,0 +1,23 @@
+name: add federalist preview link to PR
+
+on:
+  # Trigger the workflow pull request,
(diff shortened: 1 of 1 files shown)


jeremyzilar:
This is wonderful! Could we swap the base URL to `https://federalist-proxy.app.cloud.gov/preview/`?

Issue #2145, opened 2020-03-04
Bug: No visible text or indicator in Search button


hallmedia:
**Describe the bug**
Search Submit button has a hidden text node of “Search” for assistive technology, but lacks a visible text label or indicator. The CSS uses a background image of an SVG icon, however the path to the resource contains an extra ‘/’, so no image is displayed.

**Steps to Reproduce**
1. Go to https://digital.gov/communities/accessibility/
2. Inspect source of Button
     ```<button class="usa-button" name="commit" type="submit"><span class="usa-sr-only">Search</span></button>```
3. Observe the <span> is hidden
4. Observe the img src does not load via CSS

**Screenshots**
(attached)
<img width="360" alt="Screen Shot 2020-03-04 at 1 06 43 PM" src="https://user-images.githubusercontent.com/3154739/75909599-654e1c80-5e1a-11ea-8543-be5ede14bc71.png">

saracope:
Thanks!

PR #2179, opened 2020-03-11
fixing broken icons

jeremyzilar:
There are a few broken icons in our USWDS setup. This is due to the image path in the USWDS settings having a trailing slash. For example, the **search icon in the search bar**

image

Now it is `$theme-image-path: "../uswds/img";`

works! 🎉 

---

**Preview:**


saracope:
Closes #2145

PR #2186, opened 2020-03-11
Add bio for Leonardo in FSCS event

libremente:
This PR:
* adds a bio for Leonardo Favario in the Federal Source Code Summit event page. 

Let me know if you need further edits. 
Thanks! @jcastle


saracope:
Thank you, both!

PR #2241, opened 2020-03-16
New events: You Are a Product Manager (Though You May Not Know It)

ayerkes:
Creating event page for the April products in government webinar.

---

**Preview:** https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.app.cloud.gov/preview/gsa/digitalgov.gov/creating-april-products-in-government-event/event/2020/04/07/you-a-product-manager-though-you/


saracope:
@ayerkes Skippy Williams should be added as an author. 
Not sure about Lauren since she's not with an agency. @ToniBonittoGSA how do we handle those?

PR #2249, opened 2020-03-18
Fix ToS issues

### LINKED ISSUE #2233: Bug: code error breaks all PDF links on TOS page
### Description

On https://digital.gov/resources/negotiated-terms-of-service-agreements/ 
all of the PDF links are missing a space between the icon and the text. 
This may be why when you click on them, two URLs appear in the browser's addressbar.

For example, if you click on the PDF link for Let's Encrypt, browser tries to load

> https://s3.amazonaws.com/digitalgov/_legacy-img/https://letsencrypt.org/documents/LE-USG-SA-Amendment-Sept-22-2015.pdf

instead of just 

https://letsencrypt.org/documents/LE-USG-SA-Amendment-Sept-22-2015.pdf

saracope:
Changes include:
- update `letsencrypt` to `type: web` instead of .pdf. 
- fix the URI for survey analytics and wikispaces so those open correctly as well

Closes #2233

### DIFF
--- a/data/termsofservice.yml
+++ b/data/termsofservice.yml
@@ -281,7 +281,7 @@ termsofservice:
     url           : https://www.letsencrypt.org
     description   : HTTPS Certificate Authority
     uri           : 'https://letsencrypt.org/documents/LE-USG-SA-Amendment-Sept-22-2015.pdf'
-    type          : pdf
+    type          : web
     updated       :
     status        :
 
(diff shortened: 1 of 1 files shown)


saracope:
@ToniBonittoGSA I scanned through the list to fix the ones with issues. If there are still ones you find that don't work lemme know. Thanks!!

PR #2254, opened 2020-03-18
fix typo

mark-a-wilson:
Just noticed this as I was perusing your great resources.

Cheers, 
from Gov of British Columbia, Canada


saracope:
Thank you!

PR #2256, opened 2020-03-19
adding API checks in CircleCI

jeremyzilar:
We are adding in CircleCi to our builds to do test our APIs and check that they are valid JSON before going live.

**The issue** — sometimes, we are adding in new authors or topics in front-matter, and the name is misspelled, or there is some issue with the BIO that causes the JSON to break.

We'd like to know that they are going to break well before they are out in the wild.

## Here are the processes are we running

1. **Start with the HUGO** — 🙅‍♂️`hugo:latest` is not the same Hugo version that we are running locally or in federalist
2. **Install GIT** — I think that is what is happening here
```
- run: sudo apt-get update && sudo apt-get install -y --no-install-recommends git
```
3. **Checkout** — This checks out the current branch that is being built (correct?)
4. **Build Hugo** — `run: hugo -v -d $HUGO_BUILD_DIR` This builds hugo into the specified directory
5. **Echo Branch** — `command: echo ${CIRCLE_BRANCH}` Just checking which branch we are on (not needed, but has been helpful in development)
6. **Check the Authors API** — this runs `curl API-URL | jq` and checks if the API is valid or not.
```
command: curl https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.app.cloud.gov/preview/gsa/digitalgov.gov/${CIRCLE_BRANCH}/authors/v1/json/ | jq
```

---

**Preview:**


hursey013:
@saracope close in favor of https://github.com/GSA/digitalgov.gov/pull/2917?

PR #2259, opened 2020-03-20
adding sentence re: presentation date

DanielJPino-ux:

---

**Preview:**


saracope:
Here's the error code this is showing ` failed to unmarshal YAML: yaml: line 14: could not find expected ':`.
If you want to add this line in, it needs to go in the body of the page not in the front matter.
You can add the text after the `---` (triple dashes) after line 44.

PR #2282, opened 2020-03-24
updating tense language

DanielJPino-ux:
updating language to not be future tense

---

**Preview:**


DanielJPino-ux:
this is updated correctly

PR #2289, opened 2020-03-25
Setup Netlify CMS config for editing Authors Profiles

jeremyzilar:
## Goal
Setting up Netlify CMS on Digital.gov for one content type, to test the features of this integration and identify any 🚧hurdles, 🛑blockers or 🚩issues this might cause in our current workflow.

### Key Docs:
- https://federalist.18f.gov/documentation/getting-started-with-netlify-cms/
- https://www.netlifycms.org/docs/add-to-your-site/
- https://www.netlifycms.org/docs/hugo/
- https://github.com/ragasirtahk/The-ragasirtahk-Blog

## Current Issues
So far, two issues have surfaced that would affect our current workflow, and would make it more challenging to edit files directly in GitHub (website):

1. **Netlify CMS doesn't support adding coaching comments into front-matter** — (see https://github.com/netlify/netlify-cms/issues/3486) 
- > Currently, we get a fair number of edits to our sites directly through GitHub, mostly from people not in our org, but people from around the government. To make this process easier for those submitting changes, we've started adding comments in the front-matter of our pages that coach someone on how to edit a particular file (because the key:value pairs in frontmatter can be a bit hard to understand. [more »](https://github.com/netlify/netlify-cms/issues/3486)

2. **Netlify CMS doesn't support adding double quotes around field values in the front-matter** — (see https://github.com/netlify/netlify-cms/issues/3505) 
- > To make the process of editing fields easier, we are specifying that certain fields always have double quotes wrapped around them, regardless if there is a need for them or not.
- > For example, in the YML spec, the title field in the example below would not require it to be wrapped in quotes (`title: "Introduction to Accessibility"`), but because we have many editors and content designers going in to update the text fields, having double quotes around certain fields helps to ensure that the text they add will not break the build. This can be adding a colon, apostrophe, markdown formatting, etc... (`title: "NEW: Introduction to Accessibility"`). We have found that asking editors to remember to add quotes around an item if it has special chars is not something we coach them on in the workflow. [more »](https://github.com/netlify/netlify-cms/issues/3505)

## Potential Workarounds
1. We can write our own custom widgets that support the features we need.
2. We might find that once Netlify CMS is in place, that fewer people are editing our pages through Github directly.

---

## Testing: 
**Feel free to give it a try!** https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.app.cloud.gov/preview/gsa/digitalgov.gov/netlify-cms/admin/#/


jeremyzilar:
@saracope let me know if you'd like to take an hour one Friday and get this set up. I am finding it super easy ✨ to use on a few other projects.

saracope:
Thanks, @jeremyzilar! @hursey013 is going to be working on this for us and think he may be starting it this week. I'll let him know to reach out if he has some questions.

themes/digital.gov/layouts/partials/core/head.html
```diff
@@ -186,4 +186,6 @@
   <link href="https://addtocalendar.com/atc/1.5/atc-style-blue.css" rel="stylesheet" type="text/css">
   {{- end -}}
 
+  <script src="https://identity.netlify.com/v1/netlify-identity-widget.js"></script>
```
jeremyzilar, line 189: I am pretty sure you do not need this on every page. You have it already included in `themes/digital.gov/static/workflow/index.html` a that should suffice ❤️

hursey013: Hey @jeremyzilar 👋  - no you don't need it on every page, but typically you would need it on the homepage since the widget redirects to the root url when it's getting the auth token. Luckily for us it looks like Federalist provides a custom integration that takes care of the redirect instead, so we only need it on the CMS page 👍 .

hursey013:
@saracope beyond familiarizing myself with and tweaking the various configuration options, this remains largely as @jeremyzilar set it up and I believe it's ready to be tested.

The one notable addition I made was including the `slug` value within the generated front matter itself. Netlify CMS doesn't provide a way to do this through the config so I used the events generated by the CMS to hook into the save event. This isn't a perfect solution because it means the CMS is using it's own logic to create the filename slug, and we are implementing similar slugify logic to add the slug to the front matter - but because the logic is largely the same I don't believe this will be an issue.

There are some elements of Netlify CMS that I could not fully test out locally like the Editorial Workflow and deploy preview links - I also ran into some CORS issues when trying to use the CMS on the Federalist preview link, so in order to fully test this it may make sense to merge it to `main` first.

Issue #2234, opened 2020-03-13
Update the Connected Govt Act  page to use `usa-accordion card-policy`


afeijoo:
On https://digital.gov/resources/connected-government-act/, it currently includes an image of the PDF of the Act. 

Update it to use the new `usa-accordion card-policy` component for the Act (like the accordion on https://digital.gov/resources/21st-century-integrated-digital-experience-act/)

saracope:
The accordion uses `{{< card-policy >}}`.

afeijoo:
Sorry, let me try rewriting the user story to be more clear 😄 

As a visitor to https://digital.gov/resources/connected-government-act/
I expect to see an inline view of the law instead of the [current screencap image](https://s3.amazonaws.com/digitalgov/hr2331-connected-govt-act_w800.png)
So that I can read the full text of the law in HTML and find keywords in context

saracope:
@ToniBonittoGSA is going to look at it. Thanks!

ToniBonittoGSA:
https://github.com/GSA/digitalgov.gov/pull/2296

PR #2296, opened 2020-03-26
Update connected-government-act.md

ToniBonittoGSA:
add law in accordion policy card
for https://github.com/GSA/digitalgov.gov/issues/2234 

---

**Preview:** 
https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.app.cloud.gov/preview/gsa/digitalgov.gov/policy-card-for-law/resources/connected-government-act/


jeremyzilar:
@ToniBonittoGSA did you intend to have things like this in the policy card?
```
<<NOTE: Time period.>>
```
Just checking to see if you intended to highlight text, or add a shortcode here...

ToniBonittoGSA:
@jeremyzilar that's how it is in the law's copy.

thx for the highlight reminder; will take another look

jeremyzilar:
I know it is like that in the law, but maybe we should remove the << and the >> from the text to make it more readable? To my 👀 it looks like our site is broken or that HTML is not rendering. IMO, it is ok to remove things for clarity in these.

PR #2336, opened 2020-04-08
New events: Cloud.gov for Product Managers

ayerkes:
Creating cloud.gov webinar event page
---

**Preview:** https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.app.cloud.gov/preview/gsa/digitalgov.gov/creating-event-page-for-cloud.gov-webinar/event/2020/04/22/cloudgov-for-product-managers/

### DIFF
--- a/content/events/2020/04/2020-04-22-cloudgov-for-product-managers.md
+++ b/content/events/2020/04/2020-04-22-cloudgov-for-product-managers.md
@@ -33,7 +33,7 @@ A common challenge for product managers in government is deploying their applica
 
 ## Presenter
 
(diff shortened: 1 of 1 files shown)


saracope:
Needs an author page.
Bio should be shortened; would like to see less name drops there.
What's the background about why this event is getting scheduled right now?

ayerkes:
> Needs an author page.
> Bio should be shortened; would like to see less name drops there.
> What's the background about why this event is getting scheduled right now?

ayerkes:
> > Needs an author page.
> > Bio should be shortened; would like to see less name drops there.
> > What's the background about why this event is getting scheduled right now?

Hi @saracope, it has been scheduled but just hasn't been put on DG yet due to not having the event copy template from the customer and then everything changing with Zoom. 

I will remove the last sentence in her bio and add an author page.

ToniBonittoGSA:
@ayerkes author page is up; pls add her ID to this.

PR #2358, opened 2020-04-13
adding author to page

DanielJPino-ux:
adding speaker

---

**Preview:**


saracope:
Ashley Mahan should also be added as an author.
Also, when you're in there, can you fix the bio paragraphs? It looks like the text has a run-on. I think if you add 2 spaces after the first part (Matt Goodrich, Director, FedRAMP) then it should bump the rest of the text to the next line. Same for Ashley.
Screen Shot 2020-04-14 at 5 32 33 PM

PR #2382, opened 2020-04-15
updating bio

DanielJPino-ux:
adding bio for Matt Goodrich

---

**Preview:**


saracope:
Matt Goodrich left GSA so I don't think these changes are needed.

PR #2404, opened 2020-04-17
event updates

DanielJPino-ux:
updating event with speakers and updating event text

---

**Preview:**


saracope:
I think this is a duplicate PR and that event is good to go.

PR #2407, opened 2020-04-17
adding speaker

DanielJPino-ux:
adding speaker

---

**Preview:**


saracope:
Katherine Spivey doesn't seem like the right author here for a fedramp event. And it looks like that one was updated in a previous PR. Closing out.

PR #2550, opened 2020-05-06
New authors: Andres Quinones

DanielJPino-ux:
adding author to index

---

**Preview:**


DanielJPino-ux:
hey @saracope - this build shows that its passed but i get a 404 message - is it due to the special character on this person's name?

saracope:
Yeah it could be. Try taking it out and see if it builds then. @DanielJPino-ux

PR #2582, opened 2020-05-07
New authors: Robert Djurasaj

DanielJPino-ux:
adding author to index

---

**Preview:**


DanielJPino-ux:
did some investigation and the federalist wont run and i get a 404 message  - @saracope what are your thoughts? i did some investigating and found a github user with the same name: https://github.com/robertd but im not sure if that is the same person - would that impact the build?

DanielJPino-ux:
https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.app.cloud.gov/preview/gsa/digitalgov.gov/Robert-Djurasaj-Author/authors/robert-djurasaj/

PR #2646, opened 2020-05-14
Remove failing circleci config

saracope:
Removing this config to hopefully stop kicking of the build errors. This file doesn't actually do anything right now anyway.


saracope:
Also removed the webhook.

PR #2673, opened 2020-05-21
Events-Updating-June-USWDS-Call-Page

DanielJPino-ux:
deleting the registration and captioning buttons and links as those are being updated at a later time

**Summary**

* [ ] change 1

**URL / Link to page**


saracope:
It looks like the time updated also and I’m not sure it’s supposed to. I think that is a workflow tool bug??

PR #2787, opened 2020-06-17
New authors: Ravoyne "Ray" Payton

DanielJPino-ux:
adding author to index

* adding Ravoyne "Ray" Payton author to index (please keep " " for Ray)

**https://digital.gov/authors/**


DanielJPino-ux:
@saracope  or @ToniBonittoGSA -- could you look into this - i think the " " is causing a build error - should we remove the quote marks or is there another way to resolve?

saracope:
@DanielJPino-ux When you're wanting quotes inside of quotes like this you have to escape the inner quotes with a backslash in front.
EX. 
```
display_name: "Ravoyne \"Ray\" Payton"
first_name: "Ravoyne \"Ray\""
bio: "Rayvone \"Ray\" Payton ..."
```
I also added this to our documentation here:
https://github.com/GSA/digitalgov.gov/wiki/Formatting-Tips

Please let me know if that can be made more clear at all.

PR #2794, opened 2020-06-19
New sources: Section508.gov

saracope:

* adds Section508.gov as a new source
* adds link post for a page on Section508.gov

Screen Shot 2020-06-26 at 1 28 42 PM

Since this is a GSA property, I used the GSA logo we already had on the platform.


saracope:
I also added a new label for `Content: Source`, FYI @ToniBonittoGSA

PR #2874, opened 2020-07-06
Add .nvmrc file

hursey013:

The `README` recommends using a [specific version](https://github.com/GSA/digitalgov.gov#prerequisites) of Node and running the application fails when using Node 12, so adding a `.nvmrc` file will help reinforce the proper version of Node for those using nvm (and other Node version managers that support `.nvmrc`, including Federalist). This sets the version to `lts/dubnium` which is the latest release in version 10 (`v10.21.0`).


hursey013:
There is a conflict with Federalist's `hugo-build` script and `.nvmrc` files. Federalist team is looking into it and expect to have a fix in the coming days.

PR #2901, opened 2020-07-08
Update Michael Horton

afeijoo:
* Update first name 
* Add Github profile


saracope:
Also update the slug, and check to see if other pages that he's an author on need updated as well.

saracope:
Finished these updates. Ready for review.

PR #2917, opened 2020-07-09
Add automated API test to CircleCI

hursey013:

Building upon the work in https://github.com/GSA/digitalgov.gov/pull/2256, this PR adds an automated test for checking the validity of the Hugo generated JSON files used for the API. Currently it validates `authors`, `images`, and `topics` using the tool `jq`. If an issue is found with the JSON, `jq` will return a non-zero exit code causing CircleCI to fail.

Future PRs could improve the messaging behind the failures to provide more context about the problem.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,16 +1,27 @@
+version: 2.1
+
+orbs:
+  hugo: circleci/hugo@0.5.2
+
 jobs:
-  install-node:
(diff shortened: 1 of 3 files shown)


hursey013:
@saracope updated to support local testing. In your local env you should now be able to run `npm run test` (which runs the same script that CircleCI uses as well).

PR #2940, opened 2020-07-13
Use env var for go.usa.gov

hursey013:

Use environment variables for go.usa.gov API details. These env vars have been added to CircleCI, Federalist and the sample `.env` file. Based on how the counter element is implemented, it is still necessary to expose the API key to the frontend, so these changes only remove the API key from the source code and build logs but it will still be visible in the compiled HTML code.

To hide the API key from the frontend it would require the logic to be moved to the Hugo code, the drawback being the page counts will only update when new changes are committed. The other option would be to have an intermediate backend service that fetches the API data for the frontend, allowing the API key to remain only visible to the backend.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -59,8 +59,8 @@ params:
   cdnurl                : "https://s3.amazonaws.com/digitalgov"
   git_org               : "GSA"
   git_repo              : "digitalgov.gov"
-  url_shortener_login: jeremyz
-  url_shortener_api: d0ac464a2dfad044147e9537fd51503b
+  goUsaUsername         : ""
+  goUsaApiKey           : ""
   devmode               : false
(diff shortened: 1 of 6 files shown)


saracope:
@hursey013 I'm not seeing the "67 views" on local posts. It does show in the Federalist preview. I updated my .env file. Wondering if there's something else I need to do. Does that show on your local?

hursey013:
@saracope yeah that's weird, I can't get Hugo to read the `.env` file locally without explicitly passing them in through the command line like: `env HUGO_TITLE="Some Title" hugo`.

I added a workaround to get it to pass the variables in when running `npm start` - give it a shot and see if it's working for you now.

PR #2991, opened 2020-07-15
Add HTMLproofer

hursey013:

Adds https://github.com/gjtorikian/html-proofer to CI for HTML linting and dead link checking.


hursey013:
@saracope this should be ready to go. I've updated the relevant CI step to pass even though it is currently failing so we can watch the progress as the fixes are implemented. Once we get it to a point where it is no longer failing I can remove the hardcoded exit code and it'll begin to fail when future issues are found.

PR #3030, opened 2020-07-17
Make git push conditional

hursey013:

Only perform `git push` on CI when there are changes added to the commit.


hursey013:
@saracope added an additional filter because we don't need the image processing step to re-run when the changes are merged to `main` (since they will have already ran and uploaded the files on the branch).

Issue #2355, opened 2020-04-11
Bug: Fix DOI link post image


saracope:
As seen on: https://digital.gov/topics/content/

<img width="1071" alt="Screen Shot 2020-04-11 at 2 05 36 PM" src="https://user-images.githubusercontent.com/2197515/79051366-d47b1580-7bfd-11ea-9226-1f7b07b7d2dc.png">

PR #3101, opened 2020-08-03
Adds DOI logo

hursey013:

Fixes https://github.com/GSA/digitalgov.gov/issues/2355

Adds missing logo, which should resolve some of the htmlproofer errors as well.


saracope:
@hursey013 I think this should use the DOI logo instead of that (open data) blog logo since it's a Source that we might reference in a more general context.  You can see a version of the DOI logo in the [footer](https://revenuedata.doi.gov/blog/).

Issue #3044, opened 2020-07-22
Fix content errors to pass HTMLproofer


saracope:
We recently added HTMLproofer and it's identified a ton of [things to clean up](https://app.circleci.com/pipelines/github/GSA/digitalgov.gov/5119/workflows/38dd4e8a-7678-4eff-ad02-c3eabda883c3/jobs/5776/parallel-runs/0/steps/0-102).  Pulling those into this issue to chug through them.
Please check off/comment when you implement one of the fixes!

To Do:

- [ ] - ./public/2012/11/30/socialgov-summit-highlights-accessibility-challenges/index.html

 
  *  internally linking to www.noaa.gov, which does not exist (line 0)
 
    `<a href="www.noaa.gov" target="_blank">National Oceanic and Atmospheric Administration<a/>`
 
- [ ] - ./public/2013/04/30/apis-in-government/index.html
 
  *  linking to internal hash #api-education-series.md that does not exist (line 0)
 
     `<a href="/2013/03/12/resources-and-tools-for-apis/#api-education-series.md">API Education Series<a/>`
 
  *  linking to internal hash #api-standards-template.md that does not exist (line 0)
 
     `<a href="/2013/03/12/resources-and-tools-for-apis/#api-standards-template.md">API Standards Template<a/>`
 
  *  linking to internal hash #assemble-components.md that does not exist (line 0)
 
     `<a href="/2013/03/12/how-to-get-started-with-apis/#assemble-components.md" title="How to Get Started with APIs">assembling the components<a/>`

  *  linking to internal hash #audit.md that does not exist (line 0)
 
     `<a href="/2013/03/12/how-to-get-started-with-apis/#audit.md" title="How to Get Started with APIs">auditing APIs<a/>`

  *  linking to internal hash #case-studies.md that does not exist (line 0)
 
     `<a href="/2013/03/12/resources-and-tools-for-apis/#case-studies.md">Case Studies<a/>`
 
  *  linking to internal hash #common-technical-choices.md that does not exist (line 0)
 
     `<a href="/2013/03/12/api-basics/#common-technical-choices.md" title="API Basics">API technical choices<a/>`
 
  *  linking to internal hash #examples-apis-government.md that does not exist (line 0)
 
     `<a href="/2013/03/12/resources-and-tools-for-apis/#examples-apis-government.md">Examples of APIs in Government<a/>`
 
  *  linking to internal hash #government-api-listserv.md that does not exist (line 0)
 
[...]

PR #3318, opened 2020-10-10
Fixing HTML errors

danielimmke:

* This fixes from the first item in #3044  all the way down to this one:
`internally linking to FIND?s=mobile+first.md, which does not exist (line 0)`

Hacktoberfest has changed their rules, so when merging you need to add the `hacktoberfest-accepted` label for it to count. Thanks!


saracope:
@danielimmke Thanks! I'll try to review in the morning, but went ahead and labeled it as it generally looks good.

saracope (approved):

PR #3441, opened 2020-11-23
PDF slides for Plain Language Summit

mara-goldberg:
These PDFs will be linked on the PL Summit digital.gov page

* These PDFs from the PL Summit will be linked on the digital.gov event page


mara-goldberg:
Note: There should be 7 PDFs total

PR #3562, opened 2021-01-21
Update to uswds 2.10.0

### LINKED ISSUE #2894: Bug: Fix the layout shift being caused by the top gov banner
**Describe the bug**
The gov banner loading on the page is causing a big [content layout shift](https://web.dev/cls) which is having a negative impact on usability.  It's contributing to a failing CLS core on [Google PageSpeed Insights](https://developers.google.com/speed/pagespeed/insights/?url=digital.gov&tab=desktop).  
 
[`Cumulative Layout Shift (CLS)  0.12`](https://github.com/GSA/digitalgov.gov/files/4886717/screencapture-developers-google-speed-pagespeed-insights-2020-07-07-15_06_40.pdf)

In Chrome devtools, we can see that the CLS is coming from the gov banner expanding then contracting.
Screen Shot 2020-07-07 at 3 07 35 PM

saracope:

* Update to the latest release of USWDS 2.10.0
* Fixes the gov banner layout shift flash (closes #2894)
* Update to the latest version of [uswds-gulp](https://github.com/uswds/uswds-gulp)


saracope:
OK, I figured out what’s going on with the font-weights. It turns out our previous font-weight settings weren’t getting properly set. So semibold and bold font-weights were both getting translated to 700. Now that that is fixed, bold is showing as 800 which is correct. However, that def makes a difference in some places. Like as Rhett pointed out, the subheadings on Event pages are showing a difference to us.
Examples:

OLD
Screen Shot 2021-01-21 at 3 03 35 PM

NEW
Screen Shot 2021-01-21 at 3 03 21 PM

OLD
Screen Shot 2021-01-21 at 3 02 18 PM

NEW
Screen Shot 2021-01-21 at 3 02 12 PM

OLD
Screen Shot 2021-01-21 at 3 07 50 PM


NEW
Screen Shot 2021-01-21 at 3 07 56 PM

saracope:
So our next step is to decide which, if any, of the newly bold font-weights we want to change to semibold.

ToniBonittoGSA (approved):

Issue #1954, opened 2020-02-04
Create the ability to list events by topic in a markdown page


jeremyzilar:
Re: https://github.com/GSA/digitalgov.gov/pull/1953

As an editor, I would like the ability to pull in a list of events, by topic.
This would likely be done by creating HUGO shortcodes.

### Shortcodes

**Future events:** 
`{{< events-future "devops" >}}`
This would give us the ability to list all future events with the DevOps topic on a page. It would be used on the DevOps community page, and other community pages.

**Past events:** 
`{{< events-past "devops" >}}`
This would give us the ability to list all past events with the DevOps topic on a page.
- Should we consider limiting this to the last X events with a link to the DevOps topic page?
- Is this primarily a way to watch the videos, or visit the past event pages?
- Should this only pull in events that have a VIDEO?
- Is there a video specific layout and a events specific layout?

jeremyzilar:
Still working on this

saracope:
@jeremyzilar Do you already have a branch started for this?

saracope:
Use the appropriate topic tag for a community to indicate which videos would display on the page. For example, any events tagged with `plain-language-community-of-practice` would then be displayed on the [Plain Language Community](https://digital.gov/communities/plain-language/) page. The template logic would determine if the video would display in past or future events.

saracope:
Implemented this in #3570

PR #3570, opened 2021-01-26
Ks listevents

Krithisree:
Create the ability to list events by topic in a markdown page #1954
Each section will list up to 5 events at a time

**URL / Link to page**
https://digital.gov/communities/

Screen Shot 2021-04-29 at 10 08 12 AM


themes/digital.gov/layouts/communities/single.html
```diff
@@ -131,6 +131,12 @@ <h3>Community Managers</h3>
             <div class="content">
               {{- .Content -}}
             </div>
+            {{- if in .Params.topics "community-ux" -}}
```
saracope, line 134: I don't think the topic tag should be hardcoded in this template. How could it be made more generic?
We could change the community topic tag to more closely match the community slug. Would that work?

saracope (changes requested):

saracope (approved):

PR #3611, opened 2021-02-16
Request to Archive

RhettSkelton:
FedRAMP would like to archive this page as soon as possible, since it presents outdated information. We will get additional links and pages that need updating later.

* add an `expirydate` to the post
* remove a reference to the post in another piece of content

**URL / Link to page**
https://digital.gov/2015/03/12/the-new-fedramp-gov-is-here/


saracope:
needs an `expirydate` and an alias redirect added

saracope (approved):

PR #3677, opened 2021-03-06
Add Accessibility Insights

mgifford:

* I would recommend adding accessibility insights rather than AChecker.  It's a better tool and also supports Android & Windows applications (as well as web).  It doesn't do iOS too, but there are other options for Apple.

**URL / Link to page**
https://digital.gov/guides/mobile-principles/accessibility/


ToniBonittoGSA:
This was a guide created by the MobileGov Community of Practice; will contact lead.

ToniBonittoGSA:
The lead stepped down; putting on hold to review when I get back from leave 5/3

mgifford:
I think I should have nudged this a week ago so it wouldn't have been closed. Doesn't seem like I can re-open it.

PR #3872, opened 2021-05-18
Update .nvmrc to 12

aduth:

* Changes version of Node prescribed in `.nvrmc` to a newer version, v12, which is supported by Federalist (v10 is no longer supported)


thisisdano:
Thanks @aduth but we're cranking this one up to 14: https://github.com/GSA/digitalgov.gov/pull/3873

PR #3903, opened 2021-05-25
Remove link to Accessible Joe

mgifford:
The links on this page need to be updated https://digital.gov/resources/improving-the-accessibility-of-social-media-in-government/

Sadly Joseph Karr O’Connor died and this link no ([Accessible Joe](http://accessiblejoe.com/tools/)) longer works. I thought about pointing back to the archive.org page, but really that's not useful for information that was out of date at the time of the last social media update.

Removes the link to Joseph's post and some of the other posts that are outdated.  Anything not updated since 2014 about social media can be assumed to be irrelevant. 

We're building out a resource here https://accessibility.civicactions.com/guide/social-media which we would love to get more contributions to.  Government resources like this one is valuable.

### DIFF
--- a/content/resources/improving-the-accessibility-of-social-media-in-government.md
+++ b/content/resources/improving-the-accessibility-of-social-media-in-government.md
@@ -186,13 +186,10 @@ Digital Government University (DGU) offers a series of trainings based on these 
 
 ### Other Web resources
 
-* [a11yTips](http://dboudreau.tumblr.com/)
 * [WebAIM](http://webaim.org/)
 * [Usability.gov](http://www.usability.gov/)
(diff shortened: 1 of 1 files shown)


mgifford:
Actually, lots of other links here need to change too. 

Denis Boudreau is brilliant, but not so much about updating A11yTips which hasn't been updated since 2014 https://dboudreau.tumblr.com/ this is too long from the perspective of social media. His talk at Accessing Higher Ground is from 2016, but still too old 
https://accessinghigherground.org/social-media-accessibility-and-you/

AskJAN is also a broken link https://askjan.org/bulletins/SNAPTool.htm

mgifford (commented):
dropping other out-of-date links.

ToniBonittoGSA:
Hi @mgifford -- pls edit the orig comment to add the digital.gov page that needs these edits, thx!

PR #4183, opened 2021-08-31
PPT for Fedramp Analytics Case Study part 2

Gabr1elleKING:

* Adding PPT for Fedramp Analytics Case Study part 2

**https://digital.gov/event/2021/09/01/fedramp-dap-case-study-part-2-reporting-and-decision-making/**


Gabr1elleKING:
@ToniBonittoGSA any way we can proceed without the preview link?

PR #4562, opened 2022-01-28
Bump hugo version to 0.92.0

mejiaj:

* Bump Hugo version from `0.90.0` to `0.92.0`
* Add newly allowed variables in `config.yml` based on this breaking change in [`0.91.0`](https://github.com/wowchemy/wowchemy-hugo-themes/discussions/2559#discussioncomment-1840591)


mejiaj:
When running `hugo serve` I get a few warnings.

First one is:
```bash
WARN 2022/01/28 14:26:55 markup.defaultMarkdownHandler=blackfriday is deprecated and will be removed in a future release. See https://gohugo.io//content-management/formats/#list-of-content-formats
```

I see this was attempted and reverted in 87a73a0. Should we take another look at this? What pages were affected previously?

```bash
WARN 2022/01/28 14:26:56 .Path when the page is backed by a file is deprecated and will be removed in a future release. We plan to use Path for a canonical source path and you probably want to check the source is a file. To get the current behaviour, you can use a construct simlar to the below:

  {{ $path := "" }}
  {{ with .File }}
        {{ $path = .Path }}
  {{ else }}
        {{ $path = .Path }}
  {{ end }}
```

I assume this has to do with referencing project assets.


```bash
WARN 2022/01/28 14:27:01 The "tweet" shortcode will soon require two named parameters: user and id.
```

There are a handful of these, right now I see:
https://github.com/GSA/digitalgov.gov/blob/9dee7aa71a9f0af098d458e9b7da0901394d2c3a/content/news/2016/08/2016-08-08-the-content-corner-on-the-fly-content-strategies-round-offs-back-handsprings-and-double-twisting-layouts-not-required.md?plain=1#L47-L49

Will need updating or it will break.

matthewbogdan (approved):
I was able to build and run the site with these updates; note I am running hugo 0.92.1 and the updates worked successfully

ToniBonittoGSA (approved):

Issue #3539, opened 2020-12-31
Newsletter Confirmation Message Difficult to Read


devinbalkind:
**Contribute to Digital.gov**
https://digital.gov//about/contribute/

---

I subscribed to your email newsletter and the confirmation message is black on a dark background so it's hard to read. :)

![Screen Shot 2020-12-30 at 9 52 26 PM](https://user-images.githubusercontent.com/444311/103391907-84c0df00-4ae9-11eb-9a69-1180c8d8ebc4.png)

Issue #2146, opened 2020-03-04
Bug: Edit button isn't accessible


charleshalldesign:
**Describe the bug**
The “Edit Tools” link (floating action button pattern) contains no accessible name and also cannot be reached by keyboard input due to positioning method 

**Steps to Reproduce**
1. Go to https://digital.gov/communities/accessibility/
2. Inspect Edit link in bottom right corner
     ```<a class="edit edit-open" href="#"><i class="far fa-edit" aria-hidden="true"></i></a>```
3. Observe no text node
4. Navigate entire page by keyboard
5. Observe the link cannot be focussed

**Screenshots**
(attached)
<img width="90" alt="Screen Shot 2020-03-04 at 1 07 11 PM" src="https://user-images.githubusercontent.com/3154739/75910105-4439fb80-5e1b-11ea-80c2-289c93a5e70a.png">

charleshalldesign:
Commenting to address github-actions bot.
The issue persists.
Additional info: “Accessibility node not exposed. aria-hidden applied to ancestor.”

PR #4570, opened 2022-02-02
accessibility fixes

matthewbogdan:

* Updated newsletter confirmation message white text on black background #3539 
* added text character to button to make it accessible from keyboard #2146


matthewbogdan:
Will resubmit now that I have access to federalist

PR #4788, opened 2022-03-29
fix archived White House links

ryanwoldatwork:


ToniBonittoGSA:
Temporary hold - link updates are fine, but now the original Guide doc is missing from cio.gov. 
Checking with contacts; if there's no replacement/new location, will archive page.

ToniBonittoGSA:
Temporary hold - link updates are fine, but now the original Guide doc is missing from cio.gov. 
Checking with contacts; if there's no replacement/new location, will archive page.

ToniBonittoGSA:
https://github.com/GSA/digitalgov.gov/pull/4977

PR #4894, opened 2022-05-16
AS_Video Update

Schultea:

* updating CoP page for video pros now that there are new leads 

**URL / Link to page**

https://digital.gov/communities/video-production/


Schultea:
@afeijoo this should be good to go now.

PR #5170, opened 2022-07-20
Fix bug: deck replaces summary in 4 content types

nick-mon1:
Refactor template to display summary instead of deck

This change displays the `summary` field first and if blank will display the `deck`.
Current behavior displays the `deck` if present, then the `summary` which explains why when both are used, we see the deck only.

https://github.com/GSA/digitalgov.gov/blob/c6e68a44eb627db3074541e9df296b0effa02a67/themes/digital.gov/layouts/partials/core/collection.html#L78


mejiaj (approved):
LGTM, thanks!

ToniBonittoGSA (approved):

PR #5235, opened 2022-08-02
Investigate why deck is replacing summary again

nick-mon1:
**Problem**
`deck` field displays instead of the `summary` field on news, events, resources and communities pages which requires deletion of the `deck` contents to show the summary.

Current code does the opposite which is why users would have to remove the deck contents.

https://github.com/GSA/digitalgov.gov/blob/0059b153e637d090db72b2d3293328a70834059a/themes/digital.gov/layouts/news/card-article.html#L16-L20

**Fix**
New code reverses this logic: display the `summary` first then the `deck` if `summary` is empty.

### DIFF
--- a/themes/digital.gov/layouts/events/card-event-past.html
+++ b/themes/digital.gov/layouts/events/card-event-past.html
@@ -34,10 +34,10 @@
       <h3>
         <a href="{{ .Permalink }}" title="{{ .Title | markdownify }}" rel="bookmark">{{ .Title | markdownify }}</a>
       </h3>
-      {{- if .Params.deck -}}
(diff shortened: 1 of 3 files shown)


mejiaj (commented):
With logic like this, would it still render `<div class="deck">` if both are empty? Might be an edge case, but want to make sure.

```html
      {{- if .Params.summary -}}
        <div class="deck">{{ .Params.summary | markdownify -}}</div>
      {{- else -}}
        <div class="deck">{{ .Params.deck | markdownify -}}</div>
      {{- end -}}
```

ToniBonittoGSA (approved):

PR #5258, opened 2022-08-12
Find and fix issue with gulp watch not processing images

nick-mon1:
Adds a function to be called in the `gulpWatch` function to process images while `npm start` is running.


mejiaj (approved):
I was able to add an image to `./content/images/_inbox` and have the task start uploading. I'm unable to verify upload because I don't have an AWS key.

image

PR #5271, opened 2022-08-18
Remove pa11y from dependencies to save space on build

apburnes:

* Removing the explicit `pa11y` dependency from the `package.json` and using `npx` in the `test:pa11y` script so it is not installed during the site build to save disk space.


mejiaj:
@apburnes thanks for submitting this, do you know approximately how much this saves?

I worry that as the site continues to grow this will pop up again.

drewbo:
@mejiaj I think a fresh install of `pa11y-ci` is about 250MB (including some duplicated dependencies from other modules). We're also looking into longer term solutions for this problem.

mejiaj (approved):

mejiaj:
Thanks @apburnes for the PR and @drewbo on that additional info!

ToniBonittoGSA (approved):

PR #5368, opened 2022-09-14
Review <div>'s to if they need aria-role="complementary"

nick-mon1:
### Summary

Adding `aria-role=”complementary”` to `div`'s with a `aria-label="Agency description"` will improve accessibilty for screen readers.

### Problem

ARIA Labels don’t always work well on certain elements like `div`'s where there is no role `<div class="usa-identifier__identity" aria-label="Agency description">`

### Solution

Add `role="complementary"` to `div`'s


mejiaj (commented):
Out of curiosity, do you see the same violation if you changed this to a `<section>`?

I worry that using the complementary role might conflict with aside's (sidenavs) on the site. It would make more sense to use a semantic element like section to improve readability of aria label.

Otherwise this solution looks good.

nick-mon1:
I agree with your suggestion, replacing the `div` with a `section`. Tested with axe with no issues.

PR #5417, opened 2022-09-28
Add headings to all News items

nick-mon1:
### Summary
News short links need to contain an `h3` header for screen readers.

### Problem
News 'short links' posts with the `card-elsewhere` class use the `deck` property with a markdown bold syntax within the deck that looks like a header but cannot be read by the screen reader. This causes the short links to be skipped as the reader navigates down the page.

@ToniBonittoGSA we will need to update markdown templates going forward if we do this. We should remove the **bold** markup in the deck to remove the title duplication. We could update just the short link posts within the past 2 months to make the home page work with this change.

https://github.com/GSA/digitalgov.gov/blob/0d04d41b467b27bee6c15408dfe7af0077c327cd/content/news/2022/09/2022-09-27-cxday.md?plain=1#L7

### Solution
Use an `h3` for an accessible heading.

### Notes

**Before**
Screen reader headings skip over short links.

Screen Shot 2022-09-28 at 3 25 50 PM

**After**

image

### DIFF
--- a/themes/digital.gov/layouts/news/card-elsewhere.html
+++ b/themes/digital.gov/layouts/news/card-elsewhere.html
@@ -19,6 +19,9 @@
 
     <div class="summary">
       <p>
+        <h3>
+          <a href="{{- .Permalink -}}">{{- .Title | markdownify -}}</a>
+        </h3>
         {{- if .Params.deck -}}{{- .Params.deck | markdownify | emojify -}}{{- end -}}
         {{- if $source -}}
           {{ $url := urls.Parse .Params.source_url }}
--- a/themes/digital.gov/layouts/news/list.html
+++ b/themes/digital.gov/layouts/news/list.html
@@ -13,7 +13,7 @@
 
           {{/* Deck */}}
           {{- if .Params.deck -}}
-          <div class="deck">{{- .Params.deck | markdownify -}}</div>
+          <h2 class="deck">{{- .Params.deck | markdownify -}}</h2>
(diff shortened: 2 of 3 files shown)


mejiaj (changes requested):
Still getting an error with AXE Devtools

Tested [news preview](https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.app.cloud.gov/preview/gsa/digitalgov.gov/nl-add-headings-to-short-links/news/)

image

Seems to be coming from a Hubspot form. The selector is `#newsletter-signup .usa-sign-up .hbspt-form div` and it has an H2 inside.

image

### We could try

- Changing `usa-sign-up` to a section or article 
- Changing the hubspot form header (or remove all together)
- Set the `.deck` under H1 as an H2 
image

nick-mon1:
@mejiaj Thanks for the extra catches regarding the deck. I've made the first and last suggestions. The hub spot form will be removed when we replace with GovDelivery. But until then, should I just hide it with CSS?

mejiaj:
@nick-mon1 it's already hidden via `display: none;`. We'd have to try the suggestions above, mark this as blocked until hubspot is replaced, or find another way.

https://github.com/GSA/digitalgov.gov/blob/8c606573a74eb17307a24beb53e9f3c3e803b4f4/themes/digital.gov/src/scss/new/_newsletter-signup.scss#L29-L35

ToniBonittoGSA (approved):

PR #5469, opened 2022-10-11
Adds events by CoP in frontmatter

nick-mon1:
### Summary
Added front matter tags to associate events with specific CoP's to prevent displaying events on wrong community pages.

Screen Shot 2022-10-11 at 3 10 00 PM

### Problem
Upcoming events displayed on a community page are filtered by topics, instead they should be displayed by which community is hosting or co-hosting.

### Solution
Added a new taxonomy called `events` that associates events with CoP pages.

Each CoP page has a new events front matter section:

`content/communities/user-experience.md`

```
event_cop:
  - user-experience
```

Include the new front matter on the **event** page:

`content/events/2022/10/2022-09-08-testing-123.md`

```
event_cop:
  - user-experience
```

New front matter values to be used for new events:

```
- communicators
- multilingual
- plain-language
- web-managers
- web-analytics
- social-media
- user-experience
```

Workflow now has dropdown for tagging related CoP.

Screen Shot 2022-10-12 at 3 50 03 PM


mejiaj:
@nick-mon1 if I understood the problem correctly, we want to link events to specific Communities of Practice?

It seems confusing to have an `event` tag on individual events. Why not something like `related-communities` or something that describes what it's linking to?

mejiaj (commented):
@nick-mon1 looks good. 

I see in workflow that the dropdown only accepts a single value. Is there a way via workflow to add multiple communities?

nick-mon1:
Yes, good catch! I've added the `multiple:true` setting to enable multiple values.

PR #5487, opened 2022-10-14
Bump nth-check and gulp-svg-sprite

nick-mon1:
- updated gulp-svg-sprite and nth-check dependency
- updated nth-check to 2.1.1 for gulp-svg-sprite

**URL / Link to page**


package.json
```diff
@@ -46,9 +46,11 @@
   "devDependencies": {
     "del": "^3.0.0",
     "dotenv": "^8.2.0",
+    "flat-map": "^1.0.0",
```
mejiaj, line 49: Is this related to `gulp-svg-sprite` or something else?

package.json
```diff
@@ -46,9 +46,11 @@
...
     "gulp-rename": "^2.0.0",
     "gulp-responsive": "^3.0.1",
     "gulp-s3-upload": "^1.7.3",
+    "gulp-scale-images": "^3.0.2",
```
mejiaj, line 53: Is this unrelated too?

mejiaj (commented):
@nick-mon1 it looks like none of the sprite tasks are called. I'd double-check and remove if they're not being used. It could save you a dependency by removing `gulp-svg-sprite` all together.

package.json
nick-mon1: Nope, this should be removed.

package.json
nick-mon1: Yes, this is now removed

nick-mon1:
@mejiaj running `npx gulp img` cannot find `gulp-svg-sprite` when it is removed, so we need it for cli usage. I did notice that it is not directly used in the javascript but we still need it.

mejiaj (approved):
Okay, thanks @nick-mon1 for looking into that.

PR #5511, opened 2022-10-20
Improve context for navigation landmarks

nick-mon1:
- updated markup for one banner landmark
- updated navigation roles to provide context

**URL / Link to page**

### DIFF
--- a/themes/digital.gov/layouts/partials/core/footer.html
+++ b/themes/digital.gov/layouts/partials/core/footer.html
@@ -1,7 +1,7 @@
 <section class="return-to-top">
   <div class="grid-container grid-container-desktop usa-footer__return-to-top">
     <div class="grid-row">
-      <div role="navigation" class="grid-col-auto">
+      <div role="navigation" aria-label="Return to top" class="grid-col-auto">
         <a href="#">
           <span>Return to top</span>
           <svg class="usa-icon dg-icon dg-icon--standard margin-bottom-05" aria-hidden="true" focusable="false">
@@ -16,7 +16,7 @@
 <footer class="usa-footer usa-footer--big" role="contentinfo">
   <div class="usa-footer__primary-section">
     <div class="grid-container grid-container-desktop">
-      <nav class="usa-footer__nav">
+      <nav class="usa-footer__nav" aria-label="Footer">
         <div class="grid-row grid-gap-4">
           <div class="mobile-lg:grid-col-6 tablet-lg:grid-col-3">
             <section class="usa-footer__primary-content usa-footer__primary-content--collapsible">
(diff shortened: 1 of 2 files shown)


themes/digital.gov/layouts/partials/core/footer.html
```diff
@@ -1,7 +1,7 @@
 <section class="return-to-top">
   <div class="grid-container grid-container-desktop usa-footer__return-to-top">
     <div class="grid-row">
-      <div role="navigation" class="grid-col-auto">
+      <div role="navigation" aria-label="Return to top" class="grid-col-auto">
```
mejiaj, line 4: I would follow USWDS example in the [footer component ](https://designsystem.digital.gov/components/footer/).

It's inside of the footer and the markup is simplified

```html
<div class="grid-container usa-footer__return-to-top">
    <a href="#">Return to top</a>
</div>
```

themes/digital.gov/layouts/partials/core/footer.html
```diff
@@ -16,7 +16,7 @@
 <footer class="usa-footer usa-footer--big" role="contentinfo">
   <div class="usa-footer__primary-section">
     <div class="grid-container grid-container-desktop">
-      <nav class="usa-footer__nav">
+      <nav class="usa-footer__nav" aria-label="Footer">
```
mejiaj, line 19: Let's improve context by adding the word navigation to the aria label.

```suggestion
      <nav class="usa-footer__nav" aria-label="Footer navigation">
```

themes/digital.gov/layouts/partials/core/usa-banner.html
```diff
@@ -1,6 +1,6 @@
 {{- $uswds_img_path := "/uswds/img" -}}
 <a class="usa-skipnav" href="#main-content">Skip to main content</a>
-<div class="usa-banner gov-banner">
+<section class="usa-banner gov-banner">
```
mejiaj, line 3: I thought we already updated banner markup?

mejiaj (changes requested):
Nice improvements, but I think we can align closer to USWDS recommended markup. Left some comments below.

themes/digital.gov/layouts/partials/core/footer.html
nick-mon1: Thanks for the link, I've removed the `role=navigation`.

themes/digital.gov/layouts/partials/core/footer.html
nick-mon1: I've updated the code to your suggestion

themes/digital.gov/layouts/partials/core/usa-banner.html
nick-mon1: This code shouldn't have been included. It was for another branch, sorry about that!

mejiaj (approved):

PR #5513, opened 2022-10-21
Remove sitewide alert for hurricane ian

nick-mon1:
## Summary

Remove Hurricane Ian sitewide alert and put as 2nd item on the "Popular Guides & Resources"

### DIFF
--- a/content/resources/hurricane-ian-guidance-for-u-s-government-websites-and-social-media.md
+++ b/content/resources/hurricane-ian-guidance-for-u-s-government-websites-and-social-media.md
@@ -19,7 +19,7 @@ aliases:
 # 0 -- hidden
 # 1 -- visible
 # 2 -- highlighted
-weight: 2
+weight: 4
 
 topics:
   - web-content-managers-forum
--- a/themes/digital.gov/layouts/partials/core/sitewide-alert.html
+++ b/themes/digital.gov/layouts/partials/core/sitewide-alert.html
@@ -1,3 +1,4 @@
+// Creates sitewide banner alert at top of page under the banner
 <section
   class="usa-site-alert usa-site-alert--info" aria-label="Site alert">
   <div class="usa-alert">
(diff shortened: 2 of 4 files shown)


content/resources/hurricane-ian-guidance-for-u-s-government-websites-and-social-media.md
```diff
@@ -19,7 +19,7 @@ aliases:
 # 0 -- hidden
 # 1 -- visible
 # 2 -- highlighted
-weight: 2
```
mejiaj, line 22: Just curious, is there a limit on the weight? Going by the comment I only see `0, 1, 2`.

themes/digital.gov/layouts/partials/core/sitewide-alert.html
```diff
@@ -1,3 +1,4 @@
+// Creates sitewide banner alert at top of page under the banner
```
mejiaj, line 1: In the future we should templatize this further by allowing the include to take a component variant, title, and body content. This way we can save some time in the future.

mejiaj (commented):
@nick-mon1 looks good! Just some questions. Otherwise LGTM.

content/resources/hurricane-ian-guidance-for-u-s-government-websites-and-social-media.md
nick-mon1: I played around with these values to get the ordering right as requested by Ammie.
Digital.gov convention would be to use `0, 1, 2` but any integer is allowed.

themes/digital.gov/layouts/partials/core/sitewide-alert.html
nick-mon1: I'll make a note of this and have you show me what you mean by this. I think that would be an excellent idea!

content/resources/hurricane-ian-guidance-for-u-s-government-websites-and-social-media.md
mejiaj: Okay, thanks Nick. I don't think we should stray too far from this guidance. For example, maybe not go higher than a certain weight. Just something to consider for the future.

themes/digital.gov/layouts/partials/core/sitewide-alert.html
mejiaj: Sure thing, I've created an issue with the feature request
https://github.com/GSA/digitalgov.gov/issues/5514

PR #5623, opened 2022-11-28
update registry links

ryanwoldatwork:


ToniBonittoGSA:
Thanks, @ryanwoldatwork 

- this one was missing builds so I updated the blog and event pages in other PRs
- we have other updates coming for the DG's USDR page; will include link update then
- 3 don't need updates; the 2 archived pages auto-redirect, and the copy in the button.js file is just example text

PR #5521, opened 2022-10-26
Card - a11y: Place title and content before decorative image

mejiaj:
## Description

This PR implements the following **changes:**

* **Improved experience for screen readers.** Moved image _after_ text in markup so screen readers can read content before image (still keeps same visual style as on live site). [Trello ticket - Adjust news page mark-up so that image follows H3]
* **News & Updates markup refactor.** Simplified markup and removed unnecessary grid containers for news & updates on home and landing pages.
* **Cards use semantic markup.** Cards now use `<article>` element instead of a generic `div`.

### Card elsewhere

Mobile screenshots taken at 500px.

#### Before

image

#### After

image

#### Before

Tested in 1000px.

image

#### After

Slightly increased spacing between logo and section. USWDS spacing token was used to ensure consistency. 

image

### Card article

**Before**
Tested in 500px.
image

**After**
image

**Before**
image

**After**
image

### Card event

**Before**
Taken from wayback machine url from [Oct. 17, 2022](https://web.archive.org/web/20221017000438/https://digital.gov/).
image

**After**
We can now improve separation of date from content thanks to CSS grid. The border now spans the full height.
image

**Before**

image

**After**
Spacing on date|content separator is now consistent on **both** sides.
image

### Card event past

**Before**
image

**After**
image

**Before**
image

**After**
Image was increased so that it's an even 300px, instead of 293 +/-. The opacity is slightly lower to emphasize it's a past event and I didn't want to stray too far from live styles.
image

## Additional information

**Other**
* Set shared layout styles for generic cards that you can override on more specific types (e.g. events)
* Added BEM-style classes for easier and more consistent styling.
* Improved inner spacing for both mobile and desktop.

**URL / Link to page**

Pages where card styles appear (that I'm aware of).

- [Homepage](https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.app.cloud.gov/preview/gsa/digitalgov.gov/jm-a11y-news-page-image-h3/)
- [News and Updates](https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.app.cloud.gov/preview/gsa/digitalgov.gov/jm-a11y-news-page-image-h3/news/)
- [Events](https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.app.cloud.gov/preview/gsa/digitalgov.gov/jm-a11y-news-page-image-h3/events/)
[...]

PR #5626, opened 2022-11-29
Fix content layout for services directory page

nick-mon1:
### Summary

Refactored `card-service-contact` class styles to not use grid layout. Changes from [https://github.com/GSA/digitalgov.gov/pull/5521](https://github.com/GSA/digitalgov.gov/pull/5521) break styles on the [services directory page](https://digital.gov/services/directory/).

### Problem

Styles need to use original flexbox layout. The `card-service-component` has different markup than other `card` components that now use grid layout which causes unexpected content layout shifts on https://digital.gov/services/directory/

`card` markup for `grid` layout:
```
<article class="card card-event card-event-past card-linked card-event--no-video">
  <div class="card__content">...</div>
  <div class="card__media">...</div>
  <div class="card__meta">...</div>
</article>
```

`card-service-component` markup for `flex` layout:
```
<div class="card card-service-contact" data-edit-this="services/service_18f.md">
  <div class="grid-row grid-gap-2">
    <div class="grid-col-12 tablet:grid-col-3">...</div>
    <div class="grid-col-12 tablet:grid-col-5">...</div>
    <div class="grid-col-12 tablet:grid-col-4">...</div>
  </div>
</div>
```

### Solution

Removed `card` class from the services templates and copied `card` style into `card-service-contact` to match original styles.

https://user-images.githubusercontent.com/104778659/204650118-5a1d4ea7-a034-4575-9854-4f73f18eb4dd.mov

### Steps to Recreate

1. Visit https://digital.gov/services/directory/
2. **Contact** column is now displaying data justified left starting at the beginning of the column

### DIFF
--- a/themes/digital.gov/layouts/services/card-service-contact.html
+++ b/themes/digital.gov/layouts/services/card-service-contact.html
@@ -58,7 +58,7 @@
   {{- $src := .src -}}
   {{- with .item -}}
   {{ $path := "" }}{{ with .File }}{{ $path = .Path }}{{ else }}{{ $path = .Path }}{{ end }}
-  <div class="card card-service-contact" data-edit-this="{{- $path -}}">
+  <div class="card-service-contact" data-edit-this="{{- $path -}}">
 
     <div class="grid-row grid-gap-2">
       <div class="grid-col-12 tablet:grid-col-3">
--- a/themes/digital.gov/layouts/services/card-service-contact.html
+++ b/themes/digital.gov/layouts/services/card-service-contact.html
@@ -58,10 +58,10 @@
   {{- $src := .src -}}
   {{- with .item -}}
   {{ $path := "" }}{{ with .File }}{{ $path = .Path }}{{ else }}{{ $path = .Path }}{{ end }}
-  <div class="card-service-contact" data-edit-this="{{- $path -}}">
+  <div class="card card-service-contact" data-edit-this="{{- $path -}}">
 
-    <div class="grid-row grid-gap-2">
-      <div class="grid-col-12 tablet:grid-col-3">
+    <!-- <div class="grid-row grid-gap-2"> -->
+      <div class="">
 
         <div class="copy">
           {{- template "icon-resource" dict "item" . "src" $src "href" $href -}}
@@ -74,7 +74,7 @@
         </div>
 
       </div>
-      <div class="grid-col-12 tablet:grid-col-5">
+      <div class="">
 
         <div class="copy">
           {{- if .Params.deck -}}<p class="summary">{{- .Params.deck | markdownify -}}</p>{{- else -}}<p class="summary">{{- .Params.summary | markdownify | truncate 140 -}}</p>{{- end -}}
@@ -82,7 +82,7 @@
 
       </div>
 
-      <div class="grid-col-12 tablet:grid-col-4">
+      <div class="">
 
         <div class="copy copy-contacts">
           {{ if .Params.contact }}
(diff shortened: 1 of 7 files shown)


themes/digital.gov/layouts/services/card-service-contact.html
```diff
@@ -58,7 +58,7 @@
   {{- $src := .src -}}
   {{- with .item -}}
   {{ $path := "" }}{{ with .File }}{{ $path = .Path }}{{ else }}{{ $path = .Path }}{{ end }}
-  <div class="card card-service-contact" data-edit-this="{{- $path -}}">
+  <div class="card-service-contact" data-edit-this="{{- $path -}}">
```
mejiaj, line 61: The idea behind creating a generic `card` class was to work towards a consistent "card" component with variants.

We should still work towards this by either:
1. Adding overriding `card-service-contact` variant styles
2. Removing all uses of the word "card" if it's not one

---

For #1 you'd have to remove the divs with grid classes. The CSS would actually look pretty simple with CSS Grid.

```scss
.card-service-contact {
  // …previous styles
  @include at-media("tablet-lg") {
    // You could also do something like 1fr 3fr 1fr 1fr and it'd make the description column wider.
    grid-template-columns: repeat(4, 1fr);
  }
}
```

mejiaj (changes requested):
Thanks for taking this issue on. I've added a suggestion to remove grid layout classes and simplify styles.

themes/digital.gov/layouts/services/card-service-contact.html
nick-mon1: I've removed the card classes from the services contact template. These are aren't really related to the card component since they are just a simple table row listing, not an actual card component.

themes/digital.gov/layouts/services/service-contact.html
```diff
@@ -18,21 +18,21 @@
       {{/* get the path for the icon — e.g. 18f-icon.png */}}
```
mejiaj, line 18: We should refactor this page in the future. 

- Templates was the **old** way of setting partials [hugo docs](https://gohugo.io/templates/introduction/#template).
- There's a lot of repetition on how the template is being loaded, see template calls with identical parameters on lines 21, 28, and 35.

mejiaj: Created issue https://github.com/GSA/digitalgov.gov/issues/5645.

themes/digital.gov/src/scss/new/_services-directory.scss
```diff
@@ -6,7 +6,7 @@
     @include u-margin-bottom(5);
   }
 }
-.card-service-contact-heading{
+.service-contact-heading{
   @include u-display('none');
   @include at-media('tablet') {
     @include u-display('block');
```
mejiaj, line 12: Let's run prettier formatting on this partial to use consistent quotes and add spacing before `{`.

themes/digital.gov/src/scss/new/styles.scss
```diff
@@ -157,7 +157,7 @@
 @forward 'related_services';
 @forward 'services-directory';
 @forward 'card-service';
-@forward 'card-service-contact';
+@forward 'service-contact';
 
 // Topics styles
 @forward 'topic-buttons';
```
mejiaj, line 163: Let's run prettier here too to use consistent quotes.

nick-mon1:
@mejiaj Updated the minor formatting inconsistencies.

mejiaj (approved):

PR #5692, opened 2023-01-04
Add aria-label to TOC sidebar navigation

nick-mon1:
### Summary

"In this page" aside navigation is missing an aria-label. Hugo [documentation](https://gohugo.io/getting-started/configuration-markup/#table-of-contents) does not allow for setting attributes on the navigation.
`.TableOfContents` generates `<nav id="TableOfContents">` without allowing to add attributes in the configuration.

### Solution

Add `aria-label="Table of contents"` dynamically with javascript checking if `<nav id="TableOfContents">` exists on the given page.

#### Before
Screen Shot 2023-01-04 at 4 10 15 PM

#### After
Screen Shot 2023-01-04 at 4 33 55 PM

### How to Test

1. Visit any page with "In this page" aside
    - [resources](https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.app.cloud.gov/preview/gsa/digitalgov.gov/nl-label-for-toc/resources/required-web-content-and-links/)
    - [write for us](https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.app.cloud.gov/preview/gsa/digitalgov.gov/cms/news/2015/02/2015-02-02-agency-perspectives-on-personas-use-development-and-challenges/about/contribute/)
3. Open VoiceOver and should see "Table of contents navigation"

### DIFF
--- a/themes/digital.gov/src/js/common/toc.js
+++ b/themes/digital.gov/src/js/common/toc.js
@@ -1,0 +1,6 @@
+// add aria-label to the built-in .TableOfContents variable. Cannot override from configuration file
+const tableOfContents = document.querySelector('nav#TableOfContents');
+
+if (tableOfContents) {
+  tableOfContents.ariaLabel = "Table of contents";
+}
--- a/themes/digital.gov/static/js/toc.js
+++ b/themes/digital.gov/static/js/toc.js
@@ -1,12 +1,5 @@
 jQuery(function ($) {
 
-	// add aria-label to the built-in .TableOfContents variable. Cannot override from configuration file
-	const tableOfContents = document.querySelector('nav#TableOfContents');
-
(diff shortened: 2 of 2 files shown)


mejiaj (changes requested):
I appreciate the simple fix, but taking a closer look I've discovered a bigger issue. When I do a quick search of `TableOfContents` I see three results:

- layouts/partials/core/toc.html
- layouts/partials/core/guides/guide-toc.html
- layouts/shortcodes/toc.html

We should try to the best of our ability to see if these separate components are required or if we can use a single template. We should also update classes and attributes for small components like this in the template, instead of using JS.

mejiaj:
Closing this for now. We'll address the bigger issue when we replace the custom component with USWDS in-page navigation.

PR #5694, opened 2023-01-05
Short link post fixes

nick-mon1:
### Summary

Workflow displays an error that could be related to the configuration `name` field settings with a space in the field.

**Current:**
```
    name: "short link" = error
```

**New change:**
```
    name: "shortlink"  = no error is displaued
```

### Steps to recreate

1. Log in to [workflow](https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.app.cloud.gov/preview/gsa/digitalgov.gov/nl-fix-short-link-post/workflow/#/collections/authors)
2. Create shortlink post
3. Should see successful post saved and published

### Before

workflow_short_link_error

### After

Screen Shot 2023-01-09 at 1 52 54 PM

[Preview](https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.app.cloud.gov/preview/gsa/digitalgov.gov/nl-fix-short-link-post/workflow/#/collections/shortlink)

### DIFF
--- a/themes/digital.gov/static/workflow/config.yml
+++ b/themes/digital.gov/static/workflow/config.yml
@@ -789,7 +789,7 @@ collections:
 
   # SHORT LINK =======================
   - <<: *defaults
-    name: "short link"
+    name: "shortlink"
     label: "Short Link"
     label_singular: "Short Link Post"
     description: "A short link post is a shorter blog post that links to external orgs."


mejiaj (approved):
Looking at the [Netlify docs](https://www.netlifycms.org/docs/configuration-options/#:~:text=name%20(required)%3A%20unique%20identifier%20for%20the%20collection%2C%20used%20as%20the%20key%20when%20referenced%20in%20other%20contexts%20(like%20the%20relation%20widget)) I can see this must be a unique identifier. Their [example config](https://github.com/netlify/netlify-cms/blob/master/dev-test/config.yml) also uses camel case for their name values.

Thanks for fixing this @nick-mon1

PR #5759, opened 2023-01-31
Update hugo config to ignore uploads directory

nick-mon1:
### Summary

A `markdown` file is generated when a new image/file is created, these need to be removed after the file is generated.
Updating the folder path's for both images and static files should ensure this file is deleted.

### Solution

Updated hugo `config.yml` to ignore `content/uploads/_inbox` to prevent `markdown` file generation for uploads.
Updated folder path for image and static-file upload from `content/images/_inbox` to `content/uploads/_inbox`.


mejiaj (commented):
@nick-mon1 so both files and images will be uploaded to the same directory and then be processed separately?

Is it possible to upload a file and an image at the same time without issues?

nick-mon1:
@mejiaj Yes, both are uploaded to `content/uploads/_inbox' and different sub-directories `working-images` or `working-files` depending on the type.

When a new upload is created from workflow, hugo generates a markdown file but it is later removed.
See the commit history for this old PR https://github.com/GSA/digitalgov.gov/pull/5585/commits.

The changes in the config paths should remove this markdown file, they should have been updated but we found it during testing.

I've made sure to document this process between github, s3 and hugo. Still investigating why this markdown file is generated in the first place, we should be able to turn it off.

mejiaj:
@nick-mon1 is the directory that holds these deleted files ignored in the Hugo config?

https://gohugo.io/getting-started/configuration/#ignore-content-and-data-files-when-rendering

nick-mon1:
@mejiaj Yes, that is one of the paths I updated.
The other is for each of the image and file types set in the workflow config.

mejiaj (approved):

PR #5763, opened 2023-02-01
Update a11y for secondary navigation dividers

nick-mon1:
### Summary

Secondary navigation uses a hardcoded `|` as the divider which is read in JAWS as “Link About Vertical Bar, Link Subscribe Vertical Bar… (JAWS) or “Pipe” between secondary navigation links.

Solution provided by https://github.com/uswds/uswds/pull/4963 to use an entity code for the pipe separator:

`content: "\007C" / "";`

### Question

Firefox says this is an invalid value and the override fails. If I remove `/ ""` then it works.
Any suggestions for Firefox support?

### Safari

Removing `padding-left: .25rem` on `.usa-nav__secondary-item` prevents this Safari display bug:

Screen Shot 2023-02-02 at 10 14 19 AM

I've added `margin-left: .25rem` to `.usa-nav__secondary-item::before` to maintain equal spacing for the divider.

### DIFF
--- a/themes/digital.gov/src/scss/new/_navigation.scss
+++ b/themes/digital.gov/src/scss/new/_navigation.scss
@@ -1,0 +1,12 @@
+@use "uswds-core" as *;
+
+.usa-nav__secondary-links {
+  .usa-nav__secondary-item {
+    & + .usa-nav__secondary-item::before {
+      @include at-media("desktop") {
+        padding-right: units(0.5);
+        content: "\007C"; // Code for "|" character
+      }
+    }
+  }
+}
--- a/themes/digital.gov/src/scss/new/_navigation.scss
+++ b/themes/digital.gov/src/scss/new/_navigation.scss
@@ -1,19 +1,18 @@
 @use "uswds-core" as *;
 
-.usa-nav__secondary-links {
-  .usa-nav__secondary-item {
-    @include at-media("desktop") {
-      padding-left: units(0.5);
-      & + .usa-nav__secondary-item::before {
-        background-color: color("base-lighter");
-        display: inline-block;
-        width: units("1px");
-        height: units(2);
-        content: "";
-        margin-right: units(0.5);
-        vertical-align: middle;
-        padding-right: 0;
-      }
+.usa-nav__secondary-links .usa-nav__secondary-item {
+  @include at-media("desktop") {
+    padding-left: units(0.5);
+    
+    & + .usa-nav__secondary-item::before {
+      background-color: color("base-lighter");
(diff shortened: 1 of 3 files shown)


mejiaj (commented):
Good find on the Firefox issue, let's try to use a border instead.

nick-mon1:
Updated styles with these https://github.com/uswds/uswds/pull/4963/commits/9de1d42c64b2ce86a3c60bd703ee8191ca3a77ee#r1095960947. This works across all 3 browsers but does differ in width by a small amount from the current code as noted by @mahoneycm

themes/digital.gov/src/scss/new/_navigation.scss
```diff
@@ -0,0 +1,19 @@
...
+        width: units("1px");
+        height: units(2);
+        content: "";
+        margin-right: units(0.5);
+        vertical-align: middle;
+        padding-right: 0;
+      }
+    }
+  }
+}
```
mejiaj, line 19: This is the alternate solution proposed in the PR https://github.com/uswds/uswds/pull/4963#discussion_r1095960947.

This is fine since we'll update to the USWDS version with the fix once it's released.

mejiaj: We can make these styles more readable by:
1. Simplifying selector
2. Adding indents
3. Sorting by alphabetical order

```scss
.usa-nav__secondary-links .usa-nav__secondary-item {
  @include at-media("desktop") {
    padding-left: units(0.5);
    
    & + .usa-nav__secondary-item::before {
      background-color: color("base-lighter");
      content: "";
      display: inline-block;
      height: units(2);
      margin-right: units(0.5);
      padding-right: 0;
      vertical-align: middle;
      width: units("1px");
    }
  }
}
```

nick-mon1:
@mejiaj 

When I used this code from https://github.com/uswds/uswds/pull/4963#discussion_r1095960947

```
.usa-nav__secondary-links .usa-nav__secondary-item {
  @include at-media("desktop") {
    display: inline;
    padding-left: units(0.5);

    & + .usa-nav__secondary-item::before {
      background-color: color("base-lighter");
      display: inline-block;
      width: units("1px");
      height: units(2);
      content: "";
      margin-right: units(0.5);
      vertical-align: middle;
    }
  }
}
```

I got some thicker separator lines:

Screen Shot 2023-02-06 at 10 27 40 AM

That's why my code is a little different.
I've updated the code with your 3 suggestions above.

mejiaj (approved):

PR #5783, opened 2023-02-03
Removing outdated information about GSA

bpdesigns:

*Removing outdated information about GSA. We can now pay participants for research compensation using a process developed in consultation with our general counsel. Details in [the TTS handbook](https://handbook.tts.gsa.gov/18f/how-18f-works/research-guidelines/)

**URL / Link to page**
https://digital.gov/2016/02/10/paying-incentives-for-federal-user-research/


ToniBonittoGSA:
Thanks
https://github.com/GSA/digitalgov.gov/pull/5796

PR #5804, opened 2023-02-13
a11y - Resources landing

mejiaj:
Resolves card: `DG - Resources landing a11y issues`.

* Resolves a11y error with _Landmarks must be unique_
* Adds unique aria label to main navigation [0efe6ec7e51bf949773f36aa91b76b6939b043e4]
* Adds unique aria label to Table of Contents on Resources [5db4534c9bcf6db4b2624b0f491bd078e758378b]
* Adds prettier HTML formatting [d44f41669bb800de317c350ebb6a70caaa4ba606]
* Removes unnecessary grid containers on **More** links [2b17c84ebbd8b8b13c4a4b491b202aedd6216ccf]

## How to test

- Verify main navigation has aria label
- Verify Table of Contents on main navigation has aria label
- There should be **no** visual regressions

**Main navigation** (after)

`<nav aria-label="Primary navigation" class="usa-nav">`

**Table of contents** (after)
`<nav class="toc" aria-label="Resources by topic">`

**URL / Link to page**

[Resources landing →](https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.app.cloud.gov/preview/gsa/digitalgov.gov/jm-a11y-resources-landing/resources/)

### DIFF
--- a/themes/digital.gov/layouts/partials/core/header.html
+++ b/themes/digital.gov/layouts/partials/core/header.html
@@ -17,7 +17,7 @@
       </svg> Menu
     </button>
   </div>
-  <nav role="navigation" class="usa-nav">
+  <nav aria-label="Primary navigation" class="usa-nav">
     <div class="usa-nav__inner">
       <button class="usa-nav__close">
         <img src="{{ "/uswds/img/close.svg" | relURL }}" alt="close">
--- a/themes/digital.gov/layouts/resources/list.html
+++ b/themes/digital.gov/layouts/resources/list.html
@@ -88,19 +88,15 @@
                       </div>
                     </div>
                     <footer>
-                      <div class="grid-row">
-                        <div class="grid-col-12">
-                          <p class="more">
-                            <a href="{{- .Permalink -}}">
-                              <span>More on {{ .Title -}}</span>
-                              <svg class="usa-icon dg-icon dg-icon--standard margin-bottom-05" aria-hidden="true"
-                                focusable="false">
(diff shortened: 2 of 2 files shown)


themes/digital.gov/layouts/partials/core/header.html
```diff
@@ -17,7 +17,7 @@
       </svg> Menu
     </button>
   </div>
-  <nav role="navigation" class="usa-nav">
+  <nav aria-label="Primary navigation" class="usa-nav">
```
mejiaj, line 20: This adds a unique aria label, like the current USWDS header.

nick-mon1: Should we remove the "navigation" copy in the `aria-label`? It is duplicated on VoiceOver, but maybe needed for other devices?

Screen Shot 2023-02-13 at 2 00 03 PM

nick-mon1 (changes requested):

mejiaj:
@nick-mon1 I'm inclined to leave it as-is because it's recommended on USWDS. In previous testing there was a difference in the way it reads out the second navigation to clearly mark it as a navigation element.

Does it read it twice and does it sound the same to you?

nick-mon1:
@mejiaj Sounds good, just wanted to clarify from my side if that was a standard practice. It does twice and sounds the same.  Thanks for the clarification.

nick-mon1 (approved):

PR #5814, opened 2023-02-16
Update 2014-08-18-how-to-choose-a-user-experience-technique.md

edeo:
In the original line 27 to 147 looks like it is html that is not rendering the way it should be in the page, it also looks like it is duplicating lines 148 to around 203. might be good to take out.

**URL / Link to page**


ToniBonittoGSA:
Thank you, it just needs to be fixed. Added as an Issue https://github.com/GSA/digitalgov.gov/issues/5818

Issue #5831, opened 2023-02-23
Revisit use of hugo function "shuffle"


drewbo:
There are ~2 places where the [`shuffle` function is used in the site build](https://github.com/GSA/digitalgov.gov/search?q=shuffle). Because this function can create new output with no changed input, it's a source of slower builds/prevents caching content for the whole site. Each use can possibly be changed to create a similar effect without suffering the drawbacks of slower builds:
- Sometimes the `shuffle` function is used to select a light/dark placeholder avatar for authors. Instead of shuffling, we could choose light/dark based on the even/odd parity of the place in the array. This results in deterministic behavior and a potentially pleasing visual effect. (example of current behavior: https://digital.gov/authors/)
- The function is also used to show related [communities and services on posts](https://github.com/GSA/digitalgov.gov/blob/6e024ca60ac054bc8cbfdaa5e1e6884b17b5143c/themes/digital.gov/layouts/partials/core/get_related.html#L16). The function works well as written when there are more than five related options, but if there are fewer, we shuffle the data for no reason (since all of the options will be shown). When there are five or fewer related items, we could "return early" from this function and get deterministic builds with similar visual results

Would love to hear what others think about this potential improvement cc: @mejiaj

mejiaj:
@drewbo thanks for creating the issue. I've added it to our internal backlog!

drewbo:
@mejiaj Proposed solution on this branch: https://github.com/drewbo/digitalgov.gov/tree/db-no-shuffle

PR #5845, opened 2023-03-01
Db no shuffle

drewbo:
- removes use of the hugo function `shuffle` to create more consistent (faster) builds
  - this is used for creating default author thumbnails (on many pages)
  - also used for showing related communities and services (on many pages)
- implements #5831


mejiaj (approved):
Thanks for improving this! I tested:

- [x] Related services and communities show as expected
- [x] Author placeholder images load as expected

ToniBonittoGSA (approved):

nick-mon1 (approved):

PR #5898, opened 2023-03-08
Remove touchpoints from 404 page

mejiaj:

* Updated 404 template to follow [Hugo docs](https://gohugo.io/templates/404/)
* Moved touchpoints survey from `page-data.html` template (used for github edit link) to each individual template. Not idea, but necessary since we want this button to be directly below content. 

  💡 Future enhancement could move it to base template and allow a flag to toggle per page.

**URL / Link to page**

[404 page](https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.sites.pages.cloud.gov/preview/gsa/digitalgov.gov/jm-404-remove-touchpoints/404/)

**How to test**
- [x] Touchpoints should show up on previous content templates **except** 404 page
- [x] 404 renders correctly **without** touchpoints survey


themes/digital.gov/layouts/_default/404.html
```diff
@@ -32,11 +41,8 @@ <h1>{{- .Title | markdownify -}} </h1>
               {{- .Content -}}
             </div>
 
-            {{- partial "core/page-data.html" (dict "page" . "page_type" "resource" "branch" ($.Scratch.Get "branch")) -}}
-
-          </div>
+            {{- partial "core/page-data.html" (dict "page" . "page_type" "404" "branch" ($.Scratch.Get "branch")) -}}
```
mejiaj, line 44: Passed 404 as the `page_type`, but turns out this property isn't even used per comment https://github.com/GSA/digitalgov.gov/blob/fe3932d6c132e45d858054ab01470d398a6768f6/themes/digital.gov/layouts/partials/core/page-data.html#L8

ToniBonittoGSA (approved):

PR #6316, opened 2023-07-14
Enable Read Time by Default

jacksonmoody:
## Summary

Reading time on guides feature currently requires a parameter to be set manually. Updated so that read time displays by default unless explicitly turned off. 

### Preview

[Link to Preview]()

### How To Test

1. Delete a `data/guidenav` file—the read time for that guide should be displayed
2. Delete the showReadTime parameter from a `data/guidenav` file—the read time for that guide should be displayed
3. Set the showReadTime parameter to `true` on a `data/guidenav` file—the read time for that guide should be displayed
4. Set the showReadTime parameter to `false` on a `data/guidenav` file—the read time for that guide should **not** be displayed

---

### Dev Checklist

- [ ] PR has correct labels

### DIFF
--- a/themes/digital.gov/layouts/partials/core/read-time.html
+++ b/themes/digital.gov/layouts/partials/core/read-time.html
@@ -1,21 +1,17 @@
 {{- $guide_data := index $.Site.Data.guidenav (.Params.guide) -}}
-
-{{- if and $guide_data $guide_data.showReadTime -}}
-
-  {{- $readTime := 0 -}}
-
+{{- $readTime := 0 -}}
+{{- /* Show read time by default. Only disable read time if parameter is set explicitly in guidenav file */ -}}
+{{- if or (not (and $guide_data (eq $guide_data.showReadTime false))) (not $guide_data) -}}
   {{ if isset .Params "readtime" }}
     {{- $readTime = .Params.readTime -}}
   {{- else -}}
     {{- $readTime = .ReadingTime -}}
   {{- end -}}
 
-  {{ if $readTime }}
-    {{ $readTimeMinutes := cond (gt $readTime 1) "minutes" "minute" }}
-    <p class="dg-read-time">
-      Reading time:
-      {{ $readTime }}
-      {{ $readTimeMinutes }}
-    </p>
-  {{- end -}}
+  {{ $readTimeMinutes := cond (gt $readTime 1) "minutes" "minute" }}
+  <p class="dg-read-time">
+    Reading time:
+    {{ $readTime }}
+    {{ $readTimeMinutes }}
+  </p>
 {{- end -}}
--- a/themes/digital.gov/layouts/partials/core/read-time.html
+++ b/themes/digital.gov/layouts/partials/core/read-time.html
@@ -1,6 +1,8 @@
 {{- $guide_data := index $.Site.Data.guidenav (.Params.guide) -}}
 {{- $readTime := 0 -}}
+
 {{- /* Show read time by default. Only disable read time if parameter is set explicitly in guidenav file */ -}}
+
 {{- if or (not (and $guide_data (eq $guide_data.showReadTime false))) (not $guide_data) -}}
   {{ if isset .Params "readtime" }}
     {{- $readTime = .Params.readTime -}}
--- a/data/guidenav/dap.yml
+++ b/data/guidenav/dap.yml
@@ -1,5 +1,4 @@
 showNextPrevious: true
-showReadTime: true
 nav:
 - title  : 'Overview'
   path   : '/guides/dap/_index.md'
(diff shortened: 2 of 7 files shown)


themes/digital.gov/layouts/partials/core/read-time.html
```diff
@@ -1,21 +1,19 @@
 {{- $guide_data := index $.Site.Data.guidenav (.Params.guide) -}}
+{{- $readTime := 0 -}}
 
-{{- if and $guide_data $guide_data.showReadTime -}}
-
-  {{- $readTime := 0 -}}
+{{- /* Show read time by default. Only disable read time if parameter is set explicitly in guidenav file */ -}}
 
+{{- if or (not (and $guide_data (eq $guide_data.showReadTime false))) (not $guide_data) -}}
```
mejiaj, line 6: We could improve clarity of this `if` statement by extracting the check into it's own variable, like `showReadTime`

mejiaj (approved):
Looks good, thank you. I tested the following:

- [x] Verified read time shows an all guides.
- [x] Setting `showReadTime` to false hides the component.

Added a suggestion to improve clarity, but not a blocker.

themes/digital.gov/layouts/partials/core/read-time.html
jacksonmoody: Thanks! Addressed in [68462d2](https://github.com/GSA/digitalgov.gov/pull/6316/commits/68462d2fbd47c3a8d26fb81f36c423f6698268c0)

mejiaj (approved):

nick-mon1 (changes requested):
@jacksonmoody this looks great, my only suggestion would be to update `data/guidenav/_example.yml` for reference with the with the following:

```
showNextPrevious: true
showReadTime: true
nav:
....
```

Just a quick documentation note since we have have explicitly removed the `showReadTime: true'.

nick-mon1 (changes requested):
@jacksonmoody When using the `{{- partial "core/read-time.html" . -}}` in other page types like `news`, `events`, `communities` we get an error for `{{- $guide_data := index $.Site.Data.guidenav (.Params.guide) -}}`.

We can add a conditional check on the [page type](https://gohugobrasil.netlify.app/functions/getpage/) for guides logic while still supporting `.readTime` for other content types. 

Sample code of checking for page type:

```
{{ if .Site.GetPage "section" "guides" }}
  {{ $pageType := "guides"}}
  {{ printf "%#v" $pageType }}
  {{- $guide_data := index $.Site.Data.guidenav (.Params.guide) -}}
{{ end }}
```

jacksonmoody:
> On second thought, I will create a new card in trello for this enhancement. I don't want to increase the scope if this will impact guides work.

No worries! It was an easy fix. See [4a01d86](https://github.com/GSA/digitalgov.gov/pull/6316/commits/4a01d86fd416c04111cf47e1411593ccb0acbd97)

mejiaj (dismissed):
No regressions found with newly added changes. Thank you!

nick-mon1:
@jacksonmoody When using the `{{- partial "core/read-time.html" . -}}` partial on the `news/single.html` and I get this error:

`ERROR 2023/07/24 16:56:50 render of "page" failed: execute of template failed: template: news/single.html:24:18: executing "content" at <partial "core/read-time.html" .>: error calling partial: "digitalgov.gov/themes/digital.gov/layouts/partials/core/read-time.html:8:55": execute of template failed: template: partials/core/read-time.html:8:55: executing "partials/core/read-time.html" at <$guideData.showReadTime>: can't evaluate field showReadTime in type int`

Should we pass in a variable for non-guides pages?

jacksonmoody:
> @jacksonmoody When using the `{{- partial "core/read-time.html" . -}}` partial on the `news/single.html` and I get this error:

I couldn't replicate this error, but tried to address it anyways in [cc0b86b](https://github.com/GSA/digitalgov.gov/pull/6316/commits/cc0b86b45269c991a72dbb9f9d8951712c5c459b)

nick-mon1 (dismissed):
@jacksonmoody Interesting! I'll capture it as trello card to test later. Right now it works for its intended use on guide pages. Thanks for looking into this.

jacksonmoody:
> @jacksonmoody Interesting! I'll capture it as trello card to test later. Right now it works for its intended use on guide pages. Thanks for looking into this.

@nick-mon1  I think I resolved the issue in [5c0dbc2](https://github.com/GSA/digitalgov.gov/pull/6316/commits/5c0dbc292fc0efe6941760d1d1312ec2c29a8a87) if you want to take one last look!

mejiaj (approved):
Nice work. How I tested:

1. Confirm there are no regressions on guides.
2. Add read-time partial to a piece of content (`events/single.html`) and confirm it works.
3. Add `showReadTime` via frontmatter and toggle visibility based on value. For example, **only** enabled it on USWDS June monthly call and it didn't affect other events.

>**Note**
>2 & 3 are only to confirm the component works on different types of content.

PR #6330, opened 2023-07-19
Fixed pptx extension path for "supporting inclusive language through automation" event

nick-mon1:
## Summary

Fixed the url path for the `pptx` file, see https://github.com/GSA/digitalgov.gov/pull/6327/files for proper file path


mejiaj (approved):

mejiaj:
Confirming file works and looks correct in powerpoint.

PR #6419, opened 2023-08-04
HCD #1: Add Webpack for Compiling JS

jacksonmoody:
## Summary

Needed to implement 18F glossary component in #6420. Will allow us to use the `require()` function in JS code going forward.

### Preview

[Link to Preview](https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.sites.pages.cloud.gov/preview/gsa/digitalgov.gov/jtm-add-webpack/)

### How To Test

Test for regressions in existing Javascript functionality (USWDS components, GitHub editing, etc.)

> [!NOTE]  
> When switching back and forth between builds that use `Webpack` and builds that use `gulp-compile`, you may need to delete the local `dist` folder under `themes/digital.gov/static/dist` and re-run `npm start` in order for everything to compile correctly.
---

### Dev Checklist

- [x] PR has correct labels
- [x] A11y testing (voice over testing, meets WCAG, run axe tools)
- [x] Code is formatted properly

PR #6420, opened 2023-08-04
HCD #2: Add Glossary Component

jacksonmoody:
## Summary

Adds the [18F Glossary Component](https://github.com/18F/glossary) to all HCD guide pages. 
> [!WARNING]  
> Requires #6419 to be merged first

### Preview

[Link to Preview](https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.sites.pages.cloud.gov/preview/gsa/digitalgov.gov/jtm-add-hcd-glossary/)

### Solution
Screenshot 2023-08-04 at 4 00 14 PM

### How To Test
1. In the `_index.md` file of any guide, create a parameter called `glossary` and set it equal to `hcd-glossary.json`
2. Click the glossary link in the resulting guide—the HCD glossary should appear

> [!NOTE]  
>  ~~In future iterations, we should look into ways to display more than one glossary. Will require communication between Hugo and Javascript, and may require moving away from 18F component.~~
> 
> This was addressed in [4cc24b7](https://github.com/GSA/digitalgov.gov/pull/6420/commits/4cc24b7589bc4d0e13383e8a2ebb550ac369176e)

---

### Dev Checklist

- [x] PR has correct labels
- [x] A11y testing (voice over testing, meets WCAG, run axe tools)
- [x] Code is formatted properly


nick-mon1 (changes requested):
>Overall, great work @jacksonmoody! I wonder if we should move the glossary link into the menubar? I think it would handier if a guide is using the glossary to put it an easier to find location than at the top of the page. @mejiaj what do you think?

A few concerns with this approach. I'd hesitate to add a duplicate link and the header spacing is already tight, so this would take the space of a potential guide link.

I see it's already been added, so we can follow-up later to avoid rework.

mejiaj (changes requested):
Minor suggestions for improving a11y and code style.

jacksonmoody:
Thanks for the great feedback guys! Ultimately, I ended up refactoring a lot of the source code for the glossary component. This allowed me to address all of the styling changes that you suggested, and to address the concerns about the archived repository (I tried to bring all of the old JS code up to date). 

I also went ahead and modified the code so that every guide can support a different glossary. While this implementation does require passing in a Hugo variable using `<script>` tags, I think it's worth it in terms of increased flexibility for this component going forward!

To test, you can add a new `.json` file under the `themes/digital.gov/static/glossary` directory. Then, go to the `_index.md` file of any guide and add a `glossary` parameter in the front matter with the name of the `.json` file:

`glossary: "hcd-glossary.json"`

This should create a link to a glossary with these new terms on your chosen guide!

themes/digital.gov/src/js/glossary.js
```diff
@@ -0,0 +1,239 @@
...
+        .forEach((el) => el.setAttribute("aria-hidden", "true"));
+    }
+  }
+};
+
+// Toggle glossary visibility
+Glossary.prototype.toggle = function toggle() {
+  const method = this.isOpen ? this.hide : this.show;
+  method.apply(this);
+};
```
mejiaj, line 167: We could use a more modern method like [Element: toggleAttribute() method | MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/toggleAttribute) or [HTMLDetailsElement | MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDetailsElement/toggle_event).

themes/digital.gov/src/js/glossary.js
mejiaj, line 1: Nice work making the glossary more usable. Just a few concerns:

1. I don't believe we're using prototypes in other scripts. Are there major benefits to using it here?
2. Event listeners are all typically included in the `init()` function. If there are too many, consider grouping them (for example `click` on the parent and then checking if the target or currentTarget matches).
3. There are more modern methods since ES6 to do a lot of these things, like `matches()` for example.
4. Some areas could use another pass for simplification.

mejiaj (changes requested):
LGTM for the most part. 

Just some concerns on the now duplicate glossary link and some older methods used in the glossary script. If we could modernize it that would greatly help with readability and maintenance.

themes/digital.gov/src/js/glossary.js
```diff
@@ -0,0 +1,239 @@
...
+  termClass: "dg-glossary__term",
+  iconClass: "dg-glossary__icon",
+};
+
+/**
+ * Glossary side panel
+ * @constructor
+ */
+function Glossary() {
+  this.body = document.querySelector(selectors.glossaryID);
```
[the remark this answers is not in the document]
jacksonmoody: Refactored to remove IDs since we're moving towards using classes in the rest of the codebase

themes/digital.gov/src/js/glossary.js
```diff
@@ -0,0 +1,239 @@
...
+  const f =
+    p.matches ||
+    p.webkitMatchesSelector ||
+    p.mozMatchesSelector ||
+    p.msMatchesSelector ||
+    function selectorMatch(s) {
+      return [].indexOf.call(document.querySelectorAll(s), this) !== -1;
+    };
+  return f.call(el, selector);
+}
```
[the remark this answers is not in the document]
jacksonmoody: Nice! This is a bit of legacy code from the 18F implementation, but removed seeing as it's so well supported.

nick-mon1 (changes requested):
@jacksonmoody I've made a commit with a list of changes for smaller functions, improving documentation to functions, and some small style changes and a11y fix. Take a look and we can discuss if need be.

jacksonmoody:
Thanks for the comments, all! Based on @mejiaj's feedback and discussions with @nick-mon1, I've refactored `glossary.js` by moving away from the prototype structure, breaking up the code into smaller functions, and incorporating more modern ES6 features. I've also incorporated some accessibility improvements—take a look and let me know what you think!

themes/digital.gov/src/js/glossary.js
```diff
@@ -0,0 +1,251 @@
...
+
+  initializeEventListeners();
+}
+
+/* eslint-disable no-undef */
+if (glossaryBody) {
+  if (glossaryPath) {
+    initializeGlossary(glossaryPath);
+  }
+}
```
mejiaj, line 251: Couldn't you combine these?

```suggestion
if (glossaryBody && glossaryPath) {
  initializeGlossary(glossaryPath);
}
```

themes/digital.gov/src/js/glossary.js
```diff
@@ -0,0 +1,251 @@
...
+  toggleButtons.forEach((button) => {
+    button.addEventListener("click", () => showGlossary(button));
+  });
+
+  /**
+   * Close glossary on escape keypress
+   * @param {event} e to handle keycode lookup when "escape" is pressed
+   */
+  document.body.addEventListener("keyup", (e) => {
+    if (e.keyCode === KEYCODE_ESC) {
```
mejiaj, line 167: `keyCode` is deprecated, use `.code` instead.

```js
    if (e.code === "Escape") {
```

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

themes/digital.gov/src/js/glossary.js
```diff
@@ -0,0 +1,251 @@
...
+function handleTermClick(e) {
+  if (e.target.matches(`.${cssClasses.term}`)) {
+    const termNumber = e.target.getAttribute("aria-controls");
+    e.target.setAttribute(
+      "aria-expanded",
+      e.target.getAttribute("aria-expanded") === "true" ? "false" : "true"
+    );
+    glossaryBody
+      .querySelectorAll(`#${termNumber}, #${termNumber}-icon`)
+      .forEach((definition) => definition.toggleAttribute("hidden"));
```
mejiaj, line 150: It's generally not recommended to style based on the `hidden` attribute, unless you're using something like `display: none`.

We should instead use an `active` class that will change from the default state.

themes/digital.gov/src/js/glossary.js
```diff
@@ -0,0 +1,251 @@
...
+   * Close glossary when clicking outside of aside (and not on toggle button)
+   * @param {event} e to handle click on document body
+   */
+  document.body.addEventListener("click", (e) => {
+    const buttons = Array.from(toggleButtons);
+    if (!buttons.includes(e.target)) {
+      if (!glossaryBody.contains(e.target)) {
+        hideGlossary();
+      }
+    }
```
mejiaj, line 182: Can we simplify this for clarity? Consider combining.

mejiaj (changes requested):
Almost there. Added final comments, no issues with visual regressions.

Concerns with the lack of semantic markup in glossary and some of the javascript, but we can work on those as a follow-up. I'll create a trello card.

themes/digital.gov/layouts/_default/baseof.html
```diff
@@ -74,6 +74,21 @@
...
 
+  <!-- Set the glossary terms as a variable -->
+  <script type="text/javascript">
+    {{- $terms := .Params.glossary -}}
+    {{- if (not .IsNode) -}}
+      {{- $terms = .Parent.Params.glossary -}}
+    {{- end -}}
+    {{- $glossaryPath := (printf "%s/%s" "glossary/" $terms | absURL) -}}
+    let glossaryPath = null;
+    {{- if and ($terms) (os.FileExists (printf "%s/%s" "themes/digital.gov/static/glossary/" $terms)) -}}
```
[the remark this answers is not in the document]
jacksonmoody: The issue is all of this logic is needed to define the glossaryPath which is defined in the script tag. We could move this to the glossary template, but that would require adding another script tag in that file (which seems against the current pattern of keeping all the script tags in one place)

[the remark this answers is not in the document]
mejiaj: We can leave as-is for now, but I'd prefer to move away from the pattern of leaving section-specific JS in _all_ templates.

It makes more sense to modify the code where it lives naturally, in this case Guides. This makes it easier to modify & users aren't downloading unnecessary scripts.

mejiaj (approved):
Thanks for those changes.

I was able to successfully:
- [x] Add glossary.
- [x] Toggle clicking the button in Guide header.
- [x] Close by clicking the `X` button.
- [x] Close by clicking document body.
- [x] Close by hitting <kbd>Esc</kbd>
- [x] Tab through items.

nick-mon1 (approved):
Great refactoring work on `glossary.js` @jacksonmoody. I appreciate that!

PR #6494, opened 2023-08-15
Update GSA SSP PKI Blog Links

idmken:
We recently refreshed the idmanagement.gov website, which also include updating some URLs.

I updated our blog post to reference the new URLs. Thank you!


ToniBonittoGSA (approved):

ToniBonittoGSA:
> We recently refreshed the idmanagement.gov website, which also include updating some URLs.
> 
> I updated our blog post to reference the new URLs. Thank you!

Thanks for the updates -- I would recommend setting up redirects or aliases in case other sites link to the original URL(s), too.

idmken:
@ToniBonittoGSA will do!

PR #6610, opened 2023-09-19
Update 2023-09-12-cultivating-a-dynamic-career-in-ux.md

nick-mon1:
## Summary
Filename needs to be in lowercase, "UX" instead of "ux" was causing a file permissions error on AWS.

Should read `https://s3.amazonaws.com/digitalgov/static/cultivating-a-dynamic-ux-career.pptx` instead of `https://s3.amazonaws.com/digitalgov/static/cultivating-a-dynamic-UX-career.pptx`


ToniBonittoGSA:
fixed in other PR

PR #7278, opened 2024-02-06
USA Footer: Fix style specificity issue

clmedders:
## Summary

This PR applies an override to the footer styles. Related to a USWDS update [uswds/uswds#5676] that caused Digital footer to have a regression. 

**Before**
Screenshot 2024-02-06 at 4 16 40 PM

**After**
Screenshot 2024-02-06 at 4 14 12 PM

### Preview
[Link to Preview](https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.sites.pages.cloud.gov/preview/gsa/digitalgov.gov/cm-footer-fix/)

### Solution

Added a more specific style than USWDS of the width being 50% to 25%. This allowed for the footer to return to the expected result. 

There'll need to be follow up to migrate the current footer to match USWDS component footer.

### How To Test

1. Visit [preview link](https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.sites.pages.cloud.gov/preview/gsa/digitalgov.gov/cm-footer-fix/) and 
1. Scroll to the footer
1. Verify that is looks the same as the [most recent wayback format](https://web.archive.org/web/20240129091941/https://digital.gov/). There should be four columns on large screen sizes.

---

### Dev Checklist

- [ ] PR has correct labels

Tagging supervisor for reference - @rayestrada

### DIFF
--- a/themes/digital.gov/src/scss/new/_footer.scss
+++ b/themes/digital.gov/src/scss/new/_footer.scss
@@ -48,6 +48,7 @@ footer.usa-footer {
 
     .social-links {
       @include add-list-reset;
+      margin-top: units(1);
 
       li {
         i {
--- a/themes/digital.gov/src/scss/new/_uswds-overrides.scss
+++ b/themes/digital.gov/src/scss/new/_uswds-overrides.scss
@@ -60,7 +60,9 @@
 // overrides usa footer until 3.8.0 for uswds comes out and refactor footer into the uswds footer component.
 // turing off sylelint due to it flagging needing to be in BEM format
 /* stylelint-disable */
-.usa-footer .grid-row .mobile-lg\:grid-col-6 {
-  width: 25%;
+@media (min-width: 30em) {
+  .usa-footer .grid-row .mobile-lg\:grid-col-6 {
+      width: 25%;
+  }
 }
 /* stylelint-enable */
--- a/themes/digital.gov/src/scss/new/_uswds-overrides.scss
+++ b/themes/digital.gov/src/scss/new/_uswds-overrides.scss
@@ -57,8 +57,9 @@
   }
 }
 
-// overrides usa footer until uswds 3.8.0 comes out with a fix to refactor footer into the uswds footer component.
-// turing off sylelint due to it flagging needing to be in BEM format
+// Override USA footer until USWDS 3.8.0 is released
+// TODO: Reconfigure footer to match USWDS footer component
+// Disable stylelint BEM format check because we need the specificity.
 /* stylelint-disable */
 @include at-media("tablet-lg") {
   .usa-footer .grid-row .mobile-lg:grid-col-6 {
--- a/themes/digital.gov/src/scss/new/_uswds-overrides.scss
+++ b/themes/digital.gov/src/scss/new/_uswds-overrides.scss
@@ -62,7 +62,7 @@
 /* stylelint-disable */
 @media (min-width: 30em) {
   .usa-footer .grid-row .mobile-lg\:grid-col-6 {
-      width: 25%;
+    width: 25%;
   }
 }
 /* stylelint-enable */
--- a/themes/digital.gov/src/scss/new/_uswds-overrides.scss
+++ b/themes/digital.gov/src/scss/new/_uswds-overrides.scss
@@ -56,3 +56,11 @@
     mask-image: url("../img/svg-icons/material-icon-inventory_2.svg");
   }
 }
+
+// overrides usa footer until 3.8.0 for uswds comes out and refactor footer into the uswds footer component.
+// turing off sylelint due to it flagging needing to be in BEM format
+/* stylelint-disable */
+.usa-footer .grid-row .mobile-lg\:grid-col-6 {
+  width: 25%;
+}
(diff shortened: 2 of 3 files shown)


themes/digital.gov/src/scss/new/_uswds-overrides.scss
```diff
@@ -56,3 +56,11 @@
     mask-image: url("../img/svg-icons/material-icon-inventory_2.svg");
   }
 }
+
+// overrides usa footer until 3.8.0 for uswds comes out and refactor footer into the uswds footer component.
+// turing off sylelint due to it flagging needing to be in BEM format
```
clmedders, line 61: Stylelint was saying it needed to be BEM but I didnt see way to make that happen with this situation.

themes/digital.gov/src/scss/new/_uswds-overrides.scss
```diff
@@ -56,3 +56,13 @@
...
+/* stylelint-disable */
+@media (min-width: 30em) {
+  .usa-footer .grid-row .mobile-lg\:grid-col-6 {
```
mejiaj, line 64: Based on the markup

```html
<!-- Footer column: Half on mobile and fourths on tablet-lg. -->
<div class="mobile-lg:grid-col-6 tablet-lg:grid-col-3">
```

We should be targeting `grid-col-3` because we want half columns on mobile-lg and then one fourth columns on `tablet-lg`.

```suggestion
  .usa-footer .grid-row .mobile-lg\:grid-col-3 {
```

mejiaj (changes requested):
Thanks for editing. I've updated the PR title and description for clarity.

Added suggestions below for improving clarity.

themes/digital.gov/src/scss/new/_uswds-overrides.scss
```diff
@@ -56,3 +56,13 @@
     mask-image: url("../img/svg-icons/material-icon-inventory_2.svg");
   }
 }
+
+// overrides usa footer until 3.8.0 for uswds comes out and refactor footer into the uswds footer component.
+// turing off sylelint due to it flagging needing to be in BEM format
```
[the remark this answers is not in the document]
clmedders: thanks for the suggestion.. updates made

themes/digital.gov/src/scss/new/_uswds-overrides.scss
nick-mon1: Shouldn't the grid selector read like `.mobile-lg:grid-col-3`.  The `\` looks to be copied from the browser which adds this to escape it. See https://designsystem.digital.gov/utilities/layout-grid/

themes/digital.gov/src/scss/new/_uswds-overrides.scss
```diff
@@ -56,3 +56,13 @@
...
+/* stylelint-disable */
+@media (min-width: 30em) {
+  .usa-footer .grid-row .mobile-lg\:grid-col-6 {
+    width: 25%;
```
nick-mon1, line 65: Let's use the mixin instead as @mejiaj suggested.

nick-mon1 (changes requested):

themes/digital.gov/src/scss/new/_uswds-overrides.scss
clmedders: I believe the `\` is needed for the class naming... when it is removed it breaks and the issues is back with how the footer is looking. 

Screenshot 2024-02-08 at 10 13 28 AM

themes/digital.gov/src/scss/new/_uswds-overrides.scss
clmedders: @nick-mon1 can you point me to where i can find that mixin for width in percentage?  i was not able to find it in the repo or uswds

nick-mon1 (approved):
See 11a527a79ac3b11c49664a24a24fd22558c7e433 for spacing change. Ready to go!

themes/digital.gov/src/scss/new/_footer.scss
```diff
@@ -48,6 +48,7 @@ footer.usa-footer {
 
     .social-links {
       @include add-list-reset;
+      margin-top: units(1);
```
nick-mon1, line 51: I noticed a slight `top-margin` spacing difference on the `social-links` column.

**About**
Screen Shot 2024-02-08 at 1 12 20 PM

**Social Before**
Screen Shot 2024-02-08 at 1 12 05 PM

**Social After**
Screen Shot 2024-02-08 at 1 12 13 PM

themes/digital.gov/src/scss/new/_uswds-overrides.scss
nick-mon1: For percentages we have to use plain old css. The mix-in I meant for media query on line 64.

mejiaj (approved):
Looks good. Thank you.

I tested the following:
-  [x] References to `data-gh-edit-page` are gone.
-  [x] References to `gh-commit-info` are gone.
-  [x] No issues with buildJS task.
-  [x] No visual regressions.

PR #7563, opened 2024-05-14
Added hostname to og:url to include top level domain

### LINKED ISSUE #7168: Bug: og:url doesn't have complete top-level domain
**Describe the bug**

It looks like the og:url doesn't include top-level domain which could pose problems for indexing and sharing.

How Digital.gov displays (doesn't include top-level domain):

`<meta property="og:url" content="/resources/an-introduction-to-domain-management/" />`

Ideal display (includes top-level domain):

`<meta property="og:url" content="https://www.whitehouse.gov/briefing-room/speeches-remarks/2023/12/21/remarks-by-president-biden-after-marine-one-arrival-21/" />`

Reference: https://ogp.me/#url

**Steps to Reproduce**
1. Go to 'https://digital.gov/resources/an-introduction-to-domain-management/'
2. Click on 'View source' in browser
3. Scroll down to '`<meta property="og:url" content="/resources/an-introduction-to-domain-management/" />`'

**Screenshots**

nick-mon1:
## Summary
Adds top level host name to `og:url` meta tags. Closes #7168.

### Preview
[Link to Preview](https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.sites.pages.cloud.gov/preview/gsa/digitalgov.gov/nl-fix-og-url/)

### Solution

When viewing the `og:url` value on the live site, only the file pathname (`/communities/`) is displayed instead of `https://www.digital.gov/communities/`.

By setting a `hostName` params in the `config.yml` we can set the base host name.

> [!NOTE]
> `og:url` path will be different for each environment

**production**
`{{ .Permalink }}` will only return the file path and not the host name:
`https://digital.gov/resources/an-introduction-to-domain-management`

**cloud pages preview**
`{{ .Permalink }}` returns the preview path and branch name with the file path:
`https://digital.gov/preview/gsa/digitalgov.gov/nl-fix-og-url/resources/an-introduction-to-domain-management`

**localhost**
`{{ .Permalink }}` will return localhost:1313 with the file path:
`https://digital.gov/localhost:1313/resources/an-introduction-to-domain-management`

### How To Test

1. Visit any page
2. Open developer tools and search for `og:url`
3. Should see `https://digital.gov/preview/gsa/digitalgov.gov/nl-fix-og-url/resources/an-introduction-to-domain-management`
4. In production this will resolve to `https://digital.gov/resources/an-introduction-to-domain-management`

---

### Dev Checklist

- [x] PR has correct labels
- [x] A11y testing (voice over testing, meets WCAG, run axe tools)
- [x] Code is formatted properly


RileySeaburg (approved):
LGTM

themes/digital.gov/layouts/partials/core/head.html
```diff
@@ -117,7 +117,10 @@
     content="{{- $.Params.summary | default $.Site.Params.description | markdownify -}}"
   />
   <meta property="og:type" content="{{- $.Scratch.Get "pagetype" -}}" />
-  <meta property="og:url" content="{{- .Permalink -}}" />
+  <meta
+    property="og:url"
+    content="{{ .Site.Params.hostName }}{{- .Permalink -}}"
```
mejiaj, line 122: ## Communities

Returns `og:url` value of `https://digital.gov//localhost:1313/communities/`.

image

In Cloud Pages preview I see a link that leads to 404:

```
https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.sites.pages.cloud.gov/preview/gsa/digitalgov.gov/nl-fix-og-url/communities/
```

It should be:
```
https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.sites.pages.cloud.gov/preview/gsa/digitalgov.gov/nl-fix-og-url/communities/
```

mejiaj (changes requested):
@nick-mon1 I tested locally and in preview. Inner pages don't seem to generate the correct link.

I noticed the baseURL is `/` in 
https://github.com/GSA/digitalgov.gov/blob/0fd00197a3aa55b928889ba95b62acec0f350b2e/config.yml#L2

Changing it to `https://digital.gov/` makes permalinks work as expected and follows the guidance here - [Configure Hugo | Hugo](https://gohugo.io/getting-started/configuration/).

We should investigate why we need the `/` and if this is a legacy config option. Alternatively, we could try a workaround like making the permalink absolute.

**Example**
```diff
<meta
    property="og:url"
+    content="{{- .Permalink | absURL -}}"
  />
```

themes/digital.gov/layouts/partials/core/head.html
nick-mon1: @mejiaj Both of the URL's are the same in the examples above. For some reason I do not see `https://www.digital.gov` set on my local, still uses `localhost:1313`, could be a caching issue with your code suggestions.

1. Updated `baseURL: https://www.digital.gov` in `config.yml`
2. And set:
```html
<meta property="og:url" content="{{- .Permalink | absURL -}}"/>
```

themes/digital.gov/layouts/partials/core/head.html
```diff
@@ -117,7 +117,7 @@
     content="{{- $.Params.summary | default $.Site.Params.description | markdownify -}}"
   />
   <meta property="og:type" content="{{- $.Scratch.Get "pagetype" -}}" />
-  <meta property="og:url" content="{{- .Permalink -}}" />
+  <meta property="og:url" content="{{- .Permalink | absURL -}}" />
```
RileySeaburg, line 120: This is showing up as 

```
<meta property="og:url" content="/preview/gsa/digitalgov.gov/nl-fix-og-url/">
```

I'm expecting to see:

```
<meta property="og:url" content="https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.sites.pages.cloud.gov/preview/gsa/digitalgov.gov/nl-fix-og-url/preview/gsa/digitalgov.gov/nl-fix-og-url/">
```

To be clear, previous approval was based on me seeing what I expected to see. 
Now it's changed.

RileySeaburg (changes requested):
URL not displaying as expected.
Previous code change worked as expected.

nick-mon1:
### Explanation

1. Set the `baseURL: "https://digital.gov"` in the config.yml for local development. See [comment in hugo support](https://discourse.gohugo.io/t/solved-what-should-be-used-for-the-value-of-site-baseurl/5896/9).
	1. When running hugo locally, the `baseURL` will be replaced with `localhost:1313` for links to work
	2. When running in production, it will be replaced with `https://digital.gov`
	3. When `baseURL: "/"` or `baseURL: ""` (blank) it will not generate full paths in production which is what is happening currently 
2. Follow the [hugo convention](https://discourse.gohugo.io/t/hugo-update-baseurl-per-environment/31104) of using `{{ .Permalink | absURL }}` for setting `og:url` path.
3. Optional: set `hugo serve -b https://digital.gov` for testing/production environment if above fix doesn't work, this is the recommended suggestion for dealing with env configuration

### Next Steps

@RileySeaburg  the current URL's we are seeing are due to cloud.pages configuration, but they should resolve when we merge to main. If not, we should do step 3 above in a follow up PR if this fix doesn't work.

themes/digital.gov/layouts/partials/core/head.html
RileySeaburg: This may require additional cloud.gov config.

RileySeaburg (approved):

RileySeaburg:
> > 3. Optional: set `hugo serve -b https://digital.gov` for testing/production environment if above fix doesn't work, this is the recommended suggestion for dealing with env configuration
> 
> @nick-mon1 @RileySeaburg Cloud pages has some default [environment variables](https://cloud.gov/pages/documentation/env-vars-on-pages-builds/). Could we use that to target the preview builds?

@mejiaj Yes, we can.

@nick-mon1 not sure if you want to just *do that* or create a task and make this dependent on that task.
It might be worth creating a new task to review the environment variables in cloud.gov and optimize them.

mejiaj (approved):
Looks good and I don't see the notice bar banner locally anymore.

PR #7705, opened 2024-06-17
Fix img shortcode

lcummings12:
## Problem Statement 
The current production code does not support inline image positioning. Additionally, there is no class for centering images, resulting in inconsistent alignment and layout issues. Refactoring the image handling logic and introducing an image-center class will provide flexibility in image positioning and improve visual quality.

## Summary
I solved the problem by refactoring the img.html template to correctly handle image alignment. Specifically, I introduced an _align_  prop to allow for image-left, image-right, and image-center classes, ensuring flexible image positioning. 
`  {{- $align := .Get "align" | default "" -}}`

I also added an _inline_  prop to allow centered images to appear side by side. 
`{{- $inline := .Get "inline" | default "false" -}}
`
I then edited the [ _images.scss](url) file to implement the class for centering the image, as well as the class for inline images. 

```
.image-center {
  @include u-display("block");
  @include u-text-align("center");

  img {
    @include u-width("card-lg");
    @include u-maxw("full");
  }
}

.image-inline {
  @include u-display("inline-block");
  @include u-vertical-align("middle");
  @include u-text-align("center");
  @include u-margin("05");

  img {
    @include u-maxw("full");
    @include u-height("auto");
    @include u-display("inline-block");
  }
}
```

Additionally, I deleted the conditional that automatically assigned the image-right class if no alignment argument was passed and the image width was less than 600. Instead, I introduced an align parameter that ensures alignment classes are only applied when explicitly specified. Here the check also happens for the inline prop 

```
<div
  class="image{{- if $align }}
    image-{{ $align }}
  {{- end }}{{- if eq $inline "true" }}image-inline{{- end }}"
>
```

  ### How To Test
  To test these changes, I created a markdown file to use the hugo template.  I then placed various images of each of the relevant props to assure they were functioning as intended. I used assets already in the repo. 
  ``` 
  {{< img src="tile-4" align="center" alt="UX Summit Image" >}}
{{< img inline="true" align="center" src="tile-1" alt="UX Summit Image 1" >}}
{{< img inline="true" align="center" src="tile-2" alt="UX Summit Image 2" >}}

{{< img src="tile-5" align="left" alt="UX Summit Image" >}}

{{< img src="tile-3" align="right" alt="UX Summit Image" >}}
```
  
  ## Example Images
  
Screen Shot 2024-06-17 at 8 56 03 AM

1. create a file image-test.md in the content folder. 
2. copy the test images into it. 
3. visit http://localhost:1313/image-test/

### CODE REVIEW

**Comment by lcummings12** [comment]
@RileySeaburg

File: themes/digital.gov/layouts/shortcodes/img.html [hunk 1]
```diff
@@ -11,30 +12,30 @@
...
 
 <div
-  class="image{{- if (.Get "align") -}}
-    image-{{- .Get "align" -}}
-  {{- else }}
-    {{- if lt $imgWidth 600 -}}image-right{{- end -}}
-  {{- end -}}"
+  class="image{{- if $align }}
+    image-{{ $align }}
+  {{- end }}{{- if eq $inline "true" }}image-inline{{- end }}"
```
nick-mon1, line 22: Let's update this to use a go slice, similar to an array of strings. This will make it read clearer and fix this class spacing issue:

Screenshot 2024-06-17 at 3 01 52 PM
Screenshot 2024-06-17 at 3 01 52 PM



```go
{{ $imageClasses := slice "image" }}
{{ if $align }}
  {{ $imageClasses = $imageClasses | append (printf "image-%s" $align) }}
{{ end }}
{{ if eq $inline "true" }}
  {{ $imageClasses = $imageClasses | append "image-inline" }}
{{ end }}

<div class="{{ delimit $imageClasses " " }}">

```

themes/digital.gov/src/scss/new/_images.scss
nick-mon1, line 1: The new style looks good and centers multiple images like we need it to.
Screenshot 2024-06-17 at 4 29 45 PM

We should refactor the styles to use functions instead of mixins because mixins hide much of the styles and make it hard to understand what they are doing. See the [_note.scss](https://github.com/GSA/digitalgov.gov/blob/main/themes/digital.gov/src/scss/new/shortcodes/_note.scss) as an example.

Not all mixin's can be replaced but we should update as many as possible to use the function version. You can find them on the design tokens page of USWDS. See [spacing units as a starting point](https://designsystem.digital.gov/design-tokens/spacing-units/).

nick-mon1 (changes requested):
@lcummings12 Good work on addressing these image inconsistencies. I've made a few suggestions to enhance the current code. Lastly, we should update the `img` shortcode documentation on the [wiki](https://github.com/GSA/digitalgov.gov/wiki/Hugo-shortcodes) before we merge these changes.

nick-mon1:
@ToniBonittoGSA 

When using `align="left"` the image has `margin-top: 40px` applied versus `margin-top: 8px` when `align="right"` is used. This is current code and could cause some alignment issues when both are used. Just calling this out.

```html
{{< img src="image-1" align="left" alt="UX Summit Image 1" >}}
```

**align left**
Screenshot 2024-06-21 at 9 48 40 AM

themes/digital.gov/layouts/shortcodes/img.html
```diff
@@ -1,3 +1,22 @@
...
+  Displays an image with optional metadata
+  See wiki for more details https://github.com/GSA/digitalgov.gov/wiki/Hugo-shortcodes
+  TODO: finish documenting the other properties of this component
+
+  align - set the alignment for the image to be left (default if left blank), center, or right aligned.
+  inline - displays multiples images on the same line
+
+  @example for right aligned image
+  {{< img src="image-1" align="right" alt="UX Summit Image"
+>}}
```
nick-mon1, line 12: @lcummings12 I added some documentation for the new options and described what this component does.

themes/digital.gov/src/scss/new/_images.scss
nick-mon1, line 1: I removed the pixel size comments, we don't need them.

nick-mon1 (approved):

themes/digital.gov/layouts/shortcodes/img.html
```diff
@@ -1,3 +1,22 @@
+{{/* =======
+
+  Displays an image with optional metadata
+  See wiki for more details https://github.com/GSA/digitalgov.gov/wiki/Hugo-shortcodes
+  TODO: finish documenting the other properties of this component
```
RileySeaburg, line 5: @lcummings12 is this an *actual* TODO?

RileySeaburg (approved):
Just remove that TODO comment, if it's not necessary.

themes/digital.gov/layouts/shortcodes/img.html
ToniBonittoGSA: Hi @RileySeaburg and @lcummings12 

It can be removed.. but, any time we update Hugo short codes, our Wiki instructions here https://github.com/GSA/digitalgov.gov/wiki/Hugo-shortcodes **must** be updated as well. The section for images is the first one on the page.

 ~ Same goes for any development updates (gulp, etc.) for local installation instructions; those are in the Readme, here https://github.com/GSA/digitalgov.gov?tab=readme-ov-file#development-guide 

When there's a Trello card to make such updates, there should always be a checklist step to update the GitHub Wiki or the Readme's Development section.

I will merge, but pls add that as a step to the card if it isn't already (and if necessary, you can always convert a checklist item to be its own card).

thx!

ToniBonittoGSA:
> @ToniBonittoGSA
> 
> When using `align="left"` the image has `margin-top: 40px` applied versus `margin-top: 8px` when `align="right"` is used. This is current code and could cause some alignment issues when both are used. Just calling this out.
> 
> ```
> {{< img src="image-1" align="left" alt="UX Summit Image 1" >}}
> ```
> 
> **align left** Screenshot 2024-06-21 at 9 48 40 AM

Oh, my.. please make a Trello card in the backlog list to fix that margin, @nick-mon1 

Thx!

PR #7708, opened 2024-06-17
Update GTM to use new GA4 account

nick-mon1:
## Summary

Updates the old GA3 account to use the new GA4 account in the GTM script.


themes/digital.gov/layouts/_default/baseof.html
RileySeaburg, line 1: @ToniBonittoGSA so are we not using Google Tag Manager to pipe the Google Analytics through?

nick-mon1:
@RileySeaburg This PR might be premature since we haven't completed all of the steps for the move to GA4 from the google tag manager account. Learned this after I published this PR.

RileySeaburg:
@nick-mon1 I don't believe we need this at all.

GA4 is showing data being recorded up to the minute. 

I will share more details in slack.

PR #7761, opened 2024-07-02
Formatting changes for services/directory

ch-tyson:
- Removed unnecessary header margin, added padding to directory table
- Hide overflow to prevent text clipping

### Changes
**services/directory.html**
<details>

<summary>Removed unnecessary margin for header and hide overflow</summary>
before:

---

after:

</details>

See here for [preview](https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.sites.pages.cloud.gov/preview/gsa/digitalgov.gov/tc-services-formatting/services/directory/)

### How To Test

1. First Step
2. Second Step
3. Third Step

---

### Dev Checklist

- [ ] PR has correct labels

### DIFF
--- a/themes/digital.gov/layouts/services/directory.html
+++ b/themes/digital.gov/layouts/services/directory.html
@@ -25,7 +25,7 @@
     {{- $services := (where $services ".Params.weight" "ge" 1) -}}
 
 
-    <div class="grid-container grid-container-desktop-lg">
+    <div class="grid-container grid-container-desktop-lg directory">
       <div class="grid-row">
         <div class="grid-col-12">
           {{/* Gather all the service pages */}}
--- a/themes/digital.gov/src/scss/new/_service-contact.scss
+++ b/themes/digital.gov/src/scss/new/_service-contact.scss
@@ -7,7 +7,6 @@
   @include u-line-height("sans", 3);
   @include u-bg("white");
   @include u-border-bottom("1px", "gray-cool-10", "solid");
-  @include u-bg("white");
 
   @include at-media("tablet") {
     @include u-padding-y("105");
@@ -17,6 +16,7 @@
   .copy {
     @include u-display("flex");
     @include u-flex("align-start");
+    @include u-overflow("hidden");
 
     p {
       @include u-margin-top(0);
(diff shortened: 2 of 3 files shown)


ch-tyson:
Seems like this is related to #5626

nick-mon1 (changes requested):
@ch-tyson This formatting fix looks good, thanks. But we can also update the [`service-contact.html`](https://github.com/GSA/digitalgov.gov/blob/6924f0e30db049c80d5d3d4414435ccab8bb4729/themes/digital.gov/layouts/services/service-contact.html) file used in the `directory.html`.

This file could be improved by:

- The first 47 lines could be refactored to only call `service-contact` from 4 times to possibly 1 time
- Can we refactor the `$src` variable within the conditionals to be set at each check and then used to finally be passed into the `service-contact`
- We should move lines 49 - 125 into a separate partial since this is the old way of using templates in hugo.

ch-tyson:
Updated with styling changes w/ Toni's suggestions on the trello card. [Preview here](https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.sites.pages.cloud.gov/preview/gsa/digitalgov.gov/tc-services-formatting/services/directory/)

nick-mon1 (changes requested):
### Note
I've recombined the two templates into one, mostly because the `service-contact-template.html` is not used anywhere else. 

### Praise

Great work on simplifying the hugo logic for setting the `$href` and `$src`.

### Suggestion

We should update the styles for this section by:

- Preferring BEM naming conventions for the `service-contact` class
- Replacing the current uswds mixins with functions and [uswds tokens](https://designsystem.digital.gov/design-tokens/)


We should update the two stylesheets to read more like `themes/digital.gov/src/scss/new/shortcodes/_featured-resource.scss`

See #7705 for a comparison, @lcummings12 can provide some assistance with refactoring these styles.

We can use `service-contact` as the base class name and do something like `service-contact__copy`.

RileySeaburg (approved):
1. This solves the requested issue. 
2. This PR goes a long way in cleaning up the styles and ergonomics of the code. 
Great job!

nick-mon1 (approved):
Thanks @ch-tyson for the refactoring work, I added the `font-size: font-size("sans", "md")` function for setting the font size. Approved!

PR #7779, opened 2024-07-11
Fix analytics script with original id

nick-mon1:


RileySeaburg:
Issue is verified

image

PR #7879, opened 2024-08-21
Ignore .vscode directory

mejiaj:
## Summary

Adding `.vscode` directory and its contents to `.gitignore`. The local settings file was overriding my personal settings.


RileySeaburg (approved):
Thanks for catching this @mejiaj

PR #8046, opened 2024-10-11
Data about image available are missing some source image in s3

nick-mon1:
## Summary
Removes unused yaml files for images that are not used on the website. These were test uploads or duplicates of other images.

This is data cleanup for the migration.


RileySeaburg:
@nick-mon1 can you add some details to this description?

RileySeaburg (approved):
LGTM

PR #8205, opened 2024-11-25
[Do not merge] Created a sample page that lists out an example of all shortcodes.

mattsqd:
## Summary

Create a sample page that lists out an example of all short codes. This is not to be merged, simply a way to ensure that the new Drupal site looks similar to the Hugo site.

See the new page https://federalist-466b7d92-5da1-4208-974f-d61fd4348571.sites.pages.cloud.gov/preview/gsa/digitalgov.gov/shortcode-examples/short-codes/


mattsqd:
I've merged these changes into our API endpoint branch so all our stuff is in one branch. https://github.com/GSA/digitalgov.gov/pull/8254

PR #8347, opened 2025-01-21
rm usdr

ryanwoldatwork:


ryanwoldatwork:
@ToniBonittoGSA - seems like pretty much all the /services content can be deprecated, given https://digital.gov/services now redirects.

